PR #18, opened 2023-05-17
initial commit of layer selector control

jamesscottbrown:
Provides a layer selector like that on https://apps.london.gov.uk/planning/

It uses the same interactions of clicking on the circle toggling the checked/selection status of an entry, and clicking on the label toggling whether it is expanded to show the detail; I'm not sure whether this is completely intuitive to new users.  Some other CIU maps have different behaviour: on the [Census Explorer](https://apps.london.gov.uk/census-explorer/)'s context tab, clicking "London Borough" toggles whether it is selected rather than expanding.

It adds the ability to trigger these interactions by using tab and space/enter rather than clicking.

Apart from tweaking the appearance, there are open questions about how to handle two requirements:

    * sometimes it is necessary to add a control such as a e.g., a transparency slider for each layer
    * sometimes toggling the status of a layer should affect the status of child layers


ChrisKnightLDN:
Closing as this will need refactoring due to other progress

Issue #40, opened 2023-06-07
Utility function: download data as CSV


ChrisKnightLDN:

jamesscottbrown:
This is now implemented.

PR #42, opened 2023-06-07
Data download component

jamesscottbrown:
This is functional, but the styling of the button will need tweaking.

This addresses PR #40


ChrisKnightLDN:
I'm getting a number of typescript errors and The dependency on d3-dsv appears to be missing - bigger picture questions though: Should we have such a task specific component? could this just as easily be a button and keep the functionality external to the component. If this IS a component should the button inherit from the basic button component...?

jamesscottbrown:
> Should we have such a task specific component? 

I think so: this is a commonly-required piece of functionality. 

I'm also not sure that "downloading data as a file" is any more "task-specific" than "selecting a date" for filtering.

> could this just as easily be a button and keep the functionality external to the component. 

Putting it in a component does two things:

* allow the JavaScript code for downloading a data object as CSV or JSON so that it can be re-used rather than duplicated
* create a component with consistent visual appearance (i.e., same choice of download icon, icon aligned with any text)

I think both of these are useful.

We could extract the code that's run when the button is pressed to a separate file so that it can be imported independenly of the component, but I'm not convinced this is necessary: I think that in all our apps so far data donwloads have been triggered by button presses.

>If this IS a component should the button inherit from the basic button component...?

Yes, it shoudl use the main button component once that it finished.

PR #45, opened 2023-06-08
build: add button component

ChrisKnightLDN:


packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
-        I'm a button from UI!
-    </Button>
+	<Button {...args}>I'm a button from UI!</Button>
 </Template>
 
 <Story name="Default" />
+
+<Story name="Primary">
+	<Button style="primary">Primary Style</Button>
+</Story>
```
jamesscottbrown, line 19:
```suggestion
<Story name="Primary">
    <Button style="primary">Primary Style</Button>
    <Button style="primary" disabled>Primary Style (disabled)</Button>
</Story>
```

Show the enabled/disabled states for each variant side-by-side.

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+<Story name="Secondary">
+	<Button style="secondary">Secondary Style</Button>
+</Story>
```
jamesscottbrown, line 23:
```suggestion
<Story name="Secondary">
    <Button style="secondary">Secondary Style</Button>
    <Button style="secondary" disabled>Secondary Style (disabled)</Button>
</Story>
```

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+<Story name="Ghost">
+	<Button style="ghost">Ghost Style</Button>
+</Story>
```
jamesscottbrown, line 27:
```suggestion
<Story name="Ghost">
    <Button style="ghost">Ghost Style</Button>
    <Button style="ghost" disabled>Ghost Style (disabled)</Button>
</Story>
```

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+<Story name="Disabled" args={{ disabled: true }} />
```
jamesscottbrown, line 29:
```suggestion
<Story name="Disabled">
    <Button disabled>Default Style</Button>
    <Button style="primary" disabled>Primary Style</Button>
    <Button style="secondary" disabled>Secondary Style</Button>
    <Button style="ghost" disabled>Ghost Style</Button>
</Story>
```

Show what the disabled state looks like for every variant in this story.

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+	<Button size="xs">xs</Button>
+	<Button size="sm">sm</Button>
+	<Button size="base">base</Button>
+	<Button size="lg">lg</Button>
+	<Button size="xl">xl</Button>
+</Story>
+
+<Story name="With Icon">
+	<Button>Icon via slot<ArrowCircleDownIcon class="ml-2 w-6 h-6" aria-hidden="true" /></Button>
+</Story>
```
jamesscottbrown, line 41:
```suggestion
<Story name="With Icon">
	<div class="flex space-y-2 flex-col w-40">
		<Button>Icon via slot
			<ArrowCircleDownIcon class="ml-2 w-6 h-6" aria-hidden="true"/>
		</Button>

		<Button>
			<ArrowCircleDownIcon class="ml-2 w-6 h-6 mr-2" aria-hidden="true"/>
			Icon via slot
		</Button>

		<Button class="flex-col items-center">
			Icon via slot
			<ArrowCircleDownIcon class="ml-2 w-6 h-6 mr-2" aria-hidden="true"/>
		</Button>
	</div>
</Story>
```

More variants of icon positions

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+<Story name="Disabled" args={{ disabled: true }} />
+
+<Story name="Sizes">
+	<Button size="xs">xs</Button>
+	<Button size="sm">sm</Button>
+	<Button size="base">base</Button>
+	<Button size="lg">lg</Button>
+	<Button size="xl">xl</Button>
+</Story>
```
jamesscottbrown, line 37:
```suggestion
<Story name="Sizes">
    <div class="flex items-end space-x-2">
        <Button size="xs">xs</Button>
        <Button size="sm">sm</Button>
        <Button size="base">base</Button>
        <Button size="lg">lg</Button>
        <Button size="xl">xl</Button>
    </div>
</Story>
```

Align the bottom of the buttons.

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+<Story name="Disabled" args={{ disabled: true }} />
+
+<Story name="Sizes">
+	<Button size="xs">xs</Button>
+	<Button size="sm">sm</Button>
+	<Button size="base">base</Button>
+	<Button size="lg">lg</Button>
+	<Button size="xl">xl</Button>
```
jamesscottbrown, line 36: It looks a bit weird that the `lg` and `xl` buttons are the same height, due to having the same `line-height`:

```css
.text-lg {
  font-size: 1.125rem;
  line-height: 1.75rem;
}
```
```css
.text-xl {
  font-size: 1.25rem;
  line-height: 1.75rem;
}
```

jamesscottbrown:
The contrast for the `Secondary` button in the `disabled` state is a bit low:

image

image

packages/ui/src/lib/button/Button.svelte
```diff
@@ -1,3 +1,49 @@
...
+	class={buttonClass}
+	on:click
+	on:change
+	on:keydown
+	on:keyup
+	on:touchstart
+	on:touchend
+	on:touchcancel
+	on:mouseenter
+	on:mouseleave
```
PaulioRandall, line 46: **Just a note and a minor thing:** Event forwarding is nice but I've found it hides accessibility and A11y warnings. I'm not sure what to do about as yet.

Then again, the Svelte devs may be sorting it out (or perhaps they already have and I haven't noticed).

packages/ui/src/lib/button/Button.svelte
```diff
@@ -1,3 +1,49 @@
...
+		xl: 'text-xl'
+	};
+
+	let buttonClass: string;
+
+	$: buttonClass = classNames(
+		'px-4 py-2 inline-flex justify-center',
+		sizeClasses[size],
+		styleClasses[style],
+		$$props.disabled && 'cursor-not-allowed text-core-grey-300 border-core-grey-200 opacity-50',
```
PaulioRandall, line 29: I recommend making the _disabled_ prop explicit `export let disabled = false` so it's clear that it's part of the component interface. I think it's needed for TS and some other tools to validate usage.

PaulioRandall: It also won't work for `<a>` tag. You'd need to add CSS _`pointer-events: none;`_ to disable it (or equivalent Tailwind). Not an issue if we remove the anchors from the component.

PaulioRandall (commented):
A lot more comments than I usually give but it's early days. The feedback that affects the interface are the important ones.

packages/ui/src/lib/button/Button.svelte
```diff
@@ -1,3 +1,49 @@
...
+	const styleClasses = {
+		primary: 'bg-core-grey-500 border-b-4 border-core-blue-600 text-white hover:bg-core-grey-400',
+		secondary: 'bg-white border-b-4 border-core-blue-600 hover:bg-core-grey-100',
+		ghost: 'bg-transparent hover:bg-core-grey-100 dark:text-white dark:hover:bg-core-grey-400'
+	};
+
+	const sizeClasses = {
+		xs: 'text-xs',
+		sm: 'text-sm',
+		base: 'text-base',
```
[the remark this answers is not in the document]
ChrisKnightLDN: relates to font size, which runs xs, sm, base, lg ... as opposed to breakpoint which runs xs, sm, md, lg ... Whether that's the correct naming convention in this instance up for discussion, but the library would generally follow the correct conventions for font-sizing and sizing in general.

packages/ui/src/lib/button/Button.stories.svelte
```diff
@@ -1,16 +1,51 @@
...
+<Meta title="Ui/Button" component={Button} />
 
 <Template let:args>
-    <Button {...args}>
-        I'm a button from UI!
-    </Button>
+	<Button {...args}>I'm a button from UI!</Button>
 </Template>
 
 <Story name="Default" />
```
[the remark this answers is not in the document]
ChrisKnightLDN: Sounds like a storybook issue perhaps - works as you would expect for me...

packages/ui/src/lib/button/Button.svelte
```diff
@@ -1,3 +1,49 @@
-<button class="btn bg-core-blue-600 p-4">
-    <slot></slot>
-</button>
+<script lang="ts">
+	import classNames from 'classnames';
+
+	export let style: 'primary' | 'secondary' | 'ghost' = 'primary';
+	export let size: 'xs' | 'sm' | 'base' | 'lg' | 'xl' = 'base';
+	export let href: string | undefined = undefined;
+	export let type: 'button' | 'submit' = 'button';
```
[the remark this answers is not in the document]
jamesscottbrown: When used as a link the button component doesn't currently work corretly, as the `href` attribute isn't set on the `<a>` element that gets created.

Also, I'm not sure that creating navigation links that are visually indistinguishable from buttons that trigger actions is actually a good idea - it may be quite confusing for users.

See also https://github.com/Greater-London-Authority/ldn-viz-tools/pull/48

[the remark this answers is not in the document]
ChrisKnightLDN: See 'Start exploring the data' CTA for (valid imho) use case: https://apps.london.gov.uk/survey-of-londoners/

packages/ui/src/lib/button/Button.svelte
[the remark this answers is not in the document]
ChrisKnightLDN: Will need to discuss if we want fixed height buttons or allow font-size to dictate height - see also James's comment https://github.com/Greater-London-Authority/ldn-viz-tools/pull/45#discussion_r1222971160

packages/ui/src/lib/button/Button.svelte
```diff
@@ -1,3 +1,49 @@
...
+		xs: 'text-xs',
+		sm: 'text-sm',
+		base: 'text-base',
+		lg: 'text-lg',
+		xl: 'text-xl'
+	};
+
+	let buttonClass: string;
+
+	$: buttonClass = classNames(
```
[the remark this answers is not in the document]
ChrisKnightLDN: removed dependancy

packages/ui/src/lib/button/Button.svelte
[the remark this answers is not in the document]
PaulioRandall: It's a fair use case.

We could consider differing styling for page navigation buttons. We could try an arrow icon to imply you're taken somewhere when clicked?

PR #49, opened 2023-06-09
Add docs & clean package commands

PaulioRandall:
I'm adding these because I had trouble getting things running yesterday afternoon. These commands were useful.

`packages/charts/src/data/dataScatter.d.ts` keeps getting created when building. I assume it is meant to be, but no one has yet to pushed it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2956,6 +2956,10 @@
       "resolved": "packages/maps",
       "link": true
     },
+    "node_modules/@ldn-viz/themes": {
+      "resolved": "packages/themes",
+      "link": true
+    },
     "node_modules/@ldn-viz/ui": {
       "resolved": "packages/ui",
       "link": true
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
The `docs` command is definitely useful.

The `clean` command won't work on windows, as there is no `rm` command.

People often use [rimraf](https://www.npmjs.com/package/rimraf) to delte files from commands run with `npm` in a cross-platform way. But if this is mostly a convenience command for you, I don't know how much time is worth spending on making it cross-platform.

ChrisKnightLDN:
What was the problem that running that clean command is intended to solve? (obvs deleting those folders, but why?)

PaulioRandall:
> What was the problem that running that clean command is intended to solve? (obvs deleting those folders, but why?)

I had problems getting the Button branch running. `Introduction.mdx` warnings was one of them; I think James said they were fixed in `dev` now. Blasting away the gitignored files when switching branches solved half the issues.

If you think it won't be useful in future we can forgo the command.

PaulioRandall:
`clean` command removed. We can add properly later if needed. Ready for integration?

PR #53, opened 2023-06-13
Export css variables defining colors from tailwind theme

jamesscottbrown:
See also https://github.com/Greater-London-Authority/ldn-viz-tools/pull/52


mikeldn:
Agree.. this is valuable, and useful in an "external" or "alternative" context.. but we feel we should in the first instance try and use tailwind as intended... If for no other reason than to test / get used to that way of working.

PaulioRandall:
We can try the tailwind approach first but it's good to have this option visible here; even if it's just as a pull request.

I've been down the sass and tailwind route but yet find anything better than exposing plain CSS stylesheets for ensuring consuming apps are adaptable to whatever the future holds.

If the component library is built well then there should be very little, if any, styling needed in the consuming app (outside of customized components and layouts of course).

jamesscottbrown:
Whilst we can use Tailwind to style most of the UI, I don't think we can use it for our visualizations:

* if generating SVGs (as is most convenient of visualizations where the number of graphical marks is not large), we *could* style SVG elements with Tailwind classes. However, this makes exporting the SVGs for download less convenient, as the CSS that styles them isn't inline.

* if rendering to Canvas or WebGL, we need to specify colors by providing them as strings (or as arrays of numbers) when calling particular functions (or assigning values to canvas properties)

mikeldn:
STRONG preference (essential requirement) for SVG with inline styling... being able to export and use out charts "outside of the browser" is essential.

ChrisKnightLDN:
Exporting css variable and in-lining styles in our viz (svg) are two different issues. I've created a separate issue https://github.com/Greater-London-Authority/ldn-viz-tools/issues/63

PR #57, opened 2023-06-15
Classnames

ChrisKnightLDN:


packages/ui/src/lib/utils/classNames.ts
```diff
@@ -0,0 +1,3 @@
+export const classNames = (...classes: string[]) => {
+	return classes.filter(Boolean).join(' ');
+};
```
jamesscottbrown, line 3:
```suggestion

// classNames receives zero or more arguments as strings: any which are empty are discarded,
// then the remainder are joined together (separated by a space character) and returned as a single string

export const classNames = (...classes: string[]) => {
	return classes.filter(Boolean).join(' ');
};
```

I think a comment here would be useful: this is a short function but it's used in enough places that it needs to be easy for readers to understand the intent, and there are a few things to unpack (destructuring the arguments of a variadic function, and using the Boolean constructor for type coercion).

PR #58, opened 2023-06-15
Component intros

jamesscottbrown:
This adds a "Docs" section for the components that appears before the stories.


ChrisKnightLDN:
You have modal/ button mixed - I'm hesitant that we need these (now) I feel it might be more useful to document more thoroughly with our stories than by adding markdown as well...? feels like chore overhead. I suppose the problem is that our documentation is lacking?

PR #61, opened 2023-06-16
Button tweaks

jamesscottbrown:


ChrisKnightLDN:
I think we should probably explicitly type our types - I'd say it makes it easier to understand - and case in point - in this instance if href is an empty string it threw type errors in the stories. The fix was to allow string | undefined and explicitly typing those options makes that clearer.

ChrisKnightLDN (approved):

PR #62, opened 2023-06-16
create logo components

jamesscottbrown:
This adopts the same approach as the `@rgossiaux/svelte-heroicons` package that we use for icons: create Svelte components that are essentially raw SVG, but with ` {...$$restProps}` added. This means they can be sized by setting tailwind classes like `h-10` (as in the example stories).

Unlike the icon components in the `svelte-heroicons` package, these don't propogate cursor or keyboard events - I  think we'd want to use the logos for branding, rather than as interactive elements.


ChrisKnightLDN:
The svgs have inline styles - colours. It would be preferable to also be able to colour the svgs using the Heroicons approach where the inherit the (text/fill) color... https://tailwindcss.com/docs/fill

PaulioRandall:
I use this <svg> in .svelte approach in two websites now, worked really will (rotating the parts the SVG or parts of it have been a recurring requirement). I did make the mistake of having too many props and CSS variables for configuring from the outside. `{...$$restProps}` is **probably** the best way to go about it since usage will mimic <svg>.

Then again, does the logo have strict requirements for how must be presented? Much in the way of possible variations?

PR #75, opened 2023-06-23
Rename tabSelector to `RadioButtonGroupSolid` 

jamesscottbrown:
This renames the tabSelector to `RadioButtonGroupSolid`, makes the current appearance an optional variant, and makes the default appearance buttons with grey/blue backgrounds.

## Default

image

## Tabs variant

image


ChrisKnightLDN:
@jamesscottbrown can you check the changes - have refactored as discussed to use radio buttons that accept slots. The proposal would be to follow this pattern and api for similarly styled checkboxGroupSolid.

jamesscottbrown:
This looks ok, but the commented-out code should be deleted before it's merged.

Are you planning to add the `CheckboxGroupSolid` to this PR, or as a separate PR?

PR #76, opened 2023-06-23
allow some checkboxes in a CheckboxGroup to be initially checked

jamesscottbrown:
Previously, all checkboxes were initially unchecked, regardless of whether or not the `selectedOptions` prop was empty.


ChrisKnightLDN (approved):
lgtm

PR #78, opened 2023-06-23
Form default styles

ChrisKnightLDN:
**What this does**
This pr moves the responsibility for default, themed form styling to the config of the central ldn-theme. It exposes a number of classes eg ```.form-input``` that apply the relevant styles. This has the bonus that both our component in the ui package and native components in projects using the package can be styled using the simpler classes. It also reduces 'boilerplate' classes in our components.

No - but it does rely on the existing tailwind forms dependency. (theme/ ui package)
Yes (for docs) - a new dependency to allow Github Flavoured Markdown.

**Is it documented?**
I have updated the docs in the base ui package.

Yes


ChrisKnightLDN:
I could have - and it might be better to include a 'kitchen sink' example, which would show our version of this: https://tailwindcss-forms.vercel.app/kitchen-sink.html (it is currently only subtly different, but does have our light and dark colors) but I wasn't sure where that would live in storybook as it is not a specific component.

ChrisKnightLDN:
Reviewed in standup

PR #104, opened 2023-07-04
Edit PR template

jamesscottbrown:
As well as making some minor edits, this adds a new question to check that any newly created components are actually exported.

This is easy to overlook - we've made this mistake a few times now.


ChrisKnightLDN (approved):
This looks fine. Maybe keep an eye on answers to first question...

PR #92, opened 2023-07-04
Sidebar elements

jamesscottbrown:
**What this does**/**Why?**

This adds components for structuring the contents of sidebars, so that they are formatted more consistently.

**Does this introduce new dependencies?**

No.

**Is it tested?**

Only via examples in storybook.

**Is it documented?**

**Are light and dark themes considered?**

No.

**Have you included changeset file?**

Yes.

PR #106, opened 2023-07-05
Abstract sidebar

PaulioRandall:
Merged into [sidebar-components-for-demo](https://github.com/Greater-London-Authority/ldn-viz-tools/tree/sidebar-components-for-demo) with [sidebar-elements](https://github.com/Greater-London-Authority/ldn-viz-tools/pull/92) to show a \[near] complete example of London Viz Sidebars.

**What this does**

It is a `Sidebar.svelte` layout component. It deals with auto-sizing, positioning, and behavior of the sidebar. I.e. separates the concerns of sidebar positioning and behavior from content.

While it does not aim to deal with Sidebar content some amount of work was required for demonstration and as a basis for in-code design and implementation.

**Why?**

This separation of concerns provides the following benefits:

1. We avoid repeating ourselves unnecessarily for the parts of various sidebars that remain the same, i.e. not the content.
2. Provides consistency in positioning and generic behavior of different sidebar components.

With this design it should be easyish to create new or custom sidebars when needed while maintaining consistency in positioning and behavior.

**How?**

- The sidebar component only concerns itself with positioning and behavior.
- The sidebar will auto-size itself based on content.
- The sidebar will open and close to the right place based on the content.
- The content must be provided by the user developer.
- Optionally, the user can provide an alternative hood (toggle bar) containing the toggle.
- There are two named slots:
  - _hood_ contains the open close toggle or tab headers. This part sticks out when the sidebar is closed (hence the name hood). If it's not needed then giving it a div with 0 width and height will hide it (I can't imagine a use case).
  - _content_ is the main content container. It must be provided by the user developer.
 - You can put what you like in the slots and size them as you please as the `Sidebar.svelte` will auto-size.
 - The slotted _hood_ and _content_ elements will handle their own sizing for mobile and desktop. There is a store in `sidebarState.ts` currently called _isWideView_ which is true if in desktop mode (always use design for mobile first).

This PR also includes:
- `sidebarState.ts` which holds globally accessible state for sidebar. This allows uses to code a map point to open the sidebar, e.g. Schools Atlas.
  - `isWideView` is true if in desktop mode.
  - `isSidebarOpen` is true if the sidebar is open.
  - `sidebarAlignment` tells you where the sidebar is currently positioned, i.e. \['top', 'bottom', 'left', 'right']. This is a string so that it's easy to print and debug.
  - `sidebarTabId` is only used with tabbed sidebars and is the ID of the currently selected tab or _null_ if no tab selected.
- `uiState.ts` which is an alternative approach to accessing global UI state (I expect we will boot one of these two `sidebarState.ts` and `uiState.ts` after a little play).
- `SimpleToggle.svelte` (private) is a placeholder component that provides out of the box toggle for opening and closing the sidebar. It's design is based on common GLA sidebar toggles (See [Figma](https://www.figma.com/file/hejo4UYFpEURSnD3BsqpXL/Design-System?type=design&node-id=754-20905&mode=design&t=qC2tFANDIgm7anRT-4).).
- `SidebarContainer` is a placeholder content panel.
- `TabbedSidebar.svelte` as an encompassing Sidebar for tabbed content. it's an attempt to lift _tabbedSidebar_ into this new structure. The code is copy+paste+modify so should be compatible with existing use. It has a few internal private components which can be public if need.

**Questions raised**

- Breakpoints?
  - How many?
  - How should they differ?
  - Should we bother with intermediate and ultra wide views?
- What should the default open close toggle look like?
- How do we make component state globally accessible and where should it be held, what's easier for the user dev?
  - Component specific imports like `sidebarState.ts`?
  - Or a single global one like `uiState.ts`?
- What should everything be called? _(mostly solved now I think)_
- What positioning options should be available?
  - What should be the norm?
  - If and when is it OK to deviate from the norm?
  - 'top' (Quake mode) might conflict with mobile menus? _(I.e. Android has its main menu swipe down from the top)_
  - But some devices also have menus at the bottom? _(I.e. Android has its nav menu float at the bottom, hidden until you touch the bottom of the screen)_
  - Should the end user be able to set a preference? _(might solve this at the expense of a little more design complexity).

**Future possibilities**

- User slides the sidebar to their desired width and height. The component currently doesn't support this but easy to add.
  - Will need to consider snapping and enforce max and min sizes.

No.

**Is it tested?**

Only manually.

**Is it documented?**

There are storybook stories if that is what you mean?

No, not yet.

No, not yet.


packages/ui/src/lib/sidebar/TabbedContent.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang='ts'>
+	import type { Tab } from './sidebarState'
+
+	import { isWideView, selectedTabId } from './sidebarState'
```
jamesscottbrown, line 4:
```suggestion
	import { isWideView } from './sidebarState'
```

packages/ui/src/lib/sidebar/DefaultHood.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang='ts'>
+	import Menu from '@rgossiaux/svelte-heroicons/solid/Menu';
+	import { isSidebarOpen } from './sidebarState'
```
jamesscottbrown, line 3:
```suggestion
	import { isSidebarOpen } from './sidebarState'
```

jamesscottbrown:
>What should everything be called?


Rather than `TabbedContent`/`TabbedHood`/`TabbedHoodTab`, I think better names are:

* `TabPanel` for the thing that displays the content once a particular tab is selected
* `TabLabel` for the label and icon for a single tab (the thing that can be clicked on to select a particular tab)
* `TabList` for tge column that contains the full set of `TabLabel`s

This would be consistent with the other components, and matches the terms for tabs [suggested by the OpenUI project](https://open-ui.org/components/tabs.research.parts/#parts-and-terminology).

The default hood could be considered to be degenerate case of the `TabList` for which there is only one tab (and hence a generic open/close menu control cna be displayed in place of a `TabLabel`).


> What should the default open close toggle look like, i.e. default hood?

How about a left/right chevron or double-chevron?
For visual consistency we should dieally use an icon from the [heroicons set](https://v1.heroicons.com/) (the Svelte library we're using as a wrapepr currently only supports v1, but we can add others individually if necessary).

PaulioRandall:
Still a draft PR so I'm still working on the linting and typescript issues. Will incorporate your suggestions though.

I can't think of a reason not to use `TabPanel`, `TabList`, and `TabLabel`. Will rename them.

> The default hood could be considered to be degenerate case of the TabList for which there is only one tab (and hence a generic open/close menu control cna be displayed in place of a TabLabel).

It depends on the TBD style for toggles and tabs at the end of the day. I'm thinking we should keep them separate since they will change for different reasons (tab components should focus on tabbing). The component name also communicates intention so if you see `TabList` in the code you know this is a map with multiple tabs in the sidebar.

> How about a left/right/up/down chevron or double-chevron (depending on the sidepar positioning, and whether it is open or closed)? For visual consistency we should dieally use an icon from the [heroicons set](https://v1.heroicons.com/) (the Svelte library we're using as a wrapepr currently only supports v1, but we can add others individually if necessary).

I like that idea but should be addressed in a separate branch/PR. Here I'm trying to separate the general positioning and behavior from the precise sidebar aesthetics (icons, colors, widths, scrolling, etc).

packages/ui/src/lib/sidebar/DefaultHood.svelte
PaulioRandall: This dependency is needed. I expect the implementation of this component will change in the near future though.

PaulioRandall:
Here's a question. Will we ever have two sidebars on a single map??

Because this design assumes there will only ever be one.

jamesscottbrown:
This has been replaced by https://github.com/Greater-London-Authority/ldn-viz-tools/pull/155

PR #113, opened 2023-08-15
User messenger for discreet but noticable end user messages

PaulioRandall:
**What this does**

Adds a discreet end user messenger to the library. AKA Toast, Snackbar, or Popup.

It allows the user programmer to present short lived informational messages to the end user. The messages can be positioned out of the way and disappear after a few seconds.

It should be repositionable depending on the app but we should have a preferred/default place to show such messages. Top center? Middle right? It's a bit difficult with maps because there's already a lot UI going on. It's got to be out of the way but noticeable. It's likely that we'll want to display them a different place on mobile too.

Alternative naming:
- Popup
- DiscreetMessenger

**Why?**

Models are used to present information to end users but aren't suitable for communicating trivial yet useful details because they interrupt user activity. The Messenger component attempts to provide discreet yet noticeable user messages that only requires a few seconds of reading and don't interrupt end user actions.

**Use cases**

1. **As feedback that things are underway when the user initiates an asynchronous or background process. This will often be accompanied by a follow up message when the process completes.** E.g. Find my location (Geo-locator) is usually asynchronous and require a few seconds, even longer on poor connections. When the user clicks the Find My Location button a discreet message informs them the search is underway and they will be taken to their location as soon as the browser finds it.
2. **Show minor warnings and errors when a background process encounters difficulty**. E.g. Find My Location is unable to find the users location, preferable with a reason (timeout, no internet, not within map bounds, etc). A discreet error message should be shown to the end user so they know the state of things and can act accordingly.
3. **To give end users hints on how to use a feature they have just activated.** E.g. when turning on a map layer with non-standard interactivity features it's helpful to let end users know without disrupting their flow. They may not care or already be aware so a Model would just be annoying. It might be useful to only show the message on the first use of the feature as not to spam in-the-know users.
4. **To let end users know their slow or asynchronous action was successful.** E.g. It may not be possible to provide a success page for all forms, especially where a lonely input is actually a form that sends web API requests that can fail (Geo-coder). The Messenger component can be used to provide success and error messages.

**Related to issue**

N/A

**How?**

Adds:
- Messenger component that manages messages and display.
- Message (private) used by Messenger.
- MessageType is the type of message: \[Notice, Success, Warning, Error].
- postMessage is a function that can be used from anywhere to post messages.

No.

**Is it tested?**

Yes, it was created within Call For Sites map.

**Is it documented?**

Not yet. In progress.

No.

Not yet.


PaulioRandall:
Wrong branches!

PR #114, opened 2023-08-15
User Messenger for discreet but noticeable end user messages

PaulioRandall:
**What this does**

Adds a discreet end user messenger to the library. AKA Toast, Snackbar, or Popup.

It allows the user programmer to present short lived informational messages to the end user. The messages can be positioned out of the way and disappear after a few seconds.

It should be repositionable depending on the app but we should have a preferred/default place to show such messages. Top center? Middle right? It's a bit difficult with maps because there's already a lot UI going on. It's got to be out of the way but noticeable. It's likely that we'll want to display them a different place on mobile too.

Alternative naming:
- Popup
- DiscreetMessenger

**Why?**

Models are used to present information to end users but aren't suitable for communicating trivial yet useful details because they interrupt user activity. The Messenger component attempts to provide discreet yet noticeable user messages that only requires a few seconds of reading and don't interrupt end user actions.

**Use cases**

1. **As feedback that things are underway when the user initiates an asynchronous or background process. This will often be accompanied by a follow up message when the process completes.** E.g. Find my location (Geo-locator) is usually asynchronous and require a few seconds, even longer on poor connections. When the user clicks the Find My Location button a discreet message informs them the search is underway and they will be taken to their location as soon as the browser finds it.
2. **Show minor warnings and errors when a background process encounters difficulty**. E.g. Find My Location is unable to find the users location, preferable with a reason (timeout, no internet, not within map bounds, etc). A discreet error message should be shown to the end user so they know the state of things and can act accordingly.
3. **To give end users hints on how to use a feature they have just activated.** E.g. when turning on a map layer with non-standard interactivity features it's helpful to let end users know without disrupting their flow. They may not care or already be aware so a Model would just be annoying. It might be useful to only show the message on the first use of the feature as not to spam in-the-know users.
4. **To let end users know their slow or asynchronous action was successful.** E.g. It may not be possible to provide a success page for all forms, especially where a lonely input is actually a form that sends web API requests that can fail (Geo-coder). The Messenger component can be used to provide success and error messages.

**Related to issue**

N/A

**How?**

Adds:
- Messenger component that manages messages and display.
- Message (private) used by Messenger.
- MessageType is the type of message: \[Notice, Success, Warning, Error].
- postMessage is a function that can be used from anywhere to post messages.

No.

**Is it tested?**

Yes, it was created within Call For Sites map.

**Is it documented?**

Not yet. In progress.

No.

Not yet.


packages/ui/src/lib/util.ts
```diff
@@ -0,0 +1,25 @@
+export const newEnterKeyHandler = (handler: (e: KeyboardEvent) => any) => {
+	const ENTER_KEY = 13;
+	return (e: KeyboardEvent) => {
+		if (e.keyCode === ENTER_KEY) {
```
jamesscottbrown, line 4:
```suggestion
	return (e: KeyboardEvent) => {
		if (e.key === 'Enter') {
```

[MDN says](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) that the `keyCode` property is deprecated: use `key` instead (or `code` in cases where you really need to distinguish between physical keys rather than characters).

jamesscottbrown:
## Naming

I think we should probably use either [`Toast`](https://open-ui.org/components/toast.research/) or [`Alert`](https://open-ui.org/components/alert.research/) as the name of the component, as therse are what OpenUI uses, and we're generally trying to adopt the same terminology as them. 

## Functionality

It might occasionally be useful for some important messages to persist until they are dismissed by the user, rather than automatically closing after 4s. `TIME_TO_LIVE` could be `export`ed, and `0` could be interpreted as measning that a timeout should not be set.

## Appearance

Perhaps there should be default icons for Success/Warning/Error message types?


## Code structure

The new file `packages/ui/src/lib/util.ts` should be in the `packages/ui/src/lib/utils` directory, since that already exists.

I'd also suggest:

* splitting the file into two, to separate the key-handling and CSS-processing code
* moving `toCSS` and its dependencies to a separate file, and renaming the `toCSS` function to something like `styleObjToString`

jamesscottbrown:
Can you reformat `src/lib/util.ts` with Prettier to get rid of the lint error?

PaulioRandall:
> Can you reformat `src/lib/util.ts` with Prettier to get rid of the lint error?

OK. The whole package needs formatting. There's typescript errors too in other packages.

jamesscottbrown:
The code for this looks ok.

A changeset file needs to be added though,

There's an error that is apparently due to Prettier not understanding how to interpret `@apply` directives in the `<script>` section of a Svelte component:

```
@ldn-viz/ui:lint: /home/runner/work/ldn-viz-tools/ldn-viz-tools/packages/ui/src/lib/messenger/Messenger.stories.svelte
@ldn-viz/ui:lint:   143:50  error  Semicolon or block is expected  ParseError
```

TBH, I'm not sure how best to handle this; there might be  configuration change that can resolve it.

## Design

@ChrisKnightLDN will porbbaly have thoughts on the design. 

In Figma Chris used `color-[positive | wanring | error]` and  `color-background-[positive | wanring | error]` colors that don't quite match those in this component - e..g, 
`color-warning` is `#EB861E` and `color-background-warning` is `#FDF2E8`, but this componennt uses  `text-core-orange-500`  (`#EB861E`) and `bg-core-orange-100` (`#FBE6D0`). Once the semantic success/warning/error color definitions are in the `@ldn-viz/themes`, it would probably be better to use them in cases like this.

Currently, the message type (Notice/Success/Warning/Error) always appears as a title within the message. It might be useful to have a `title` prop that overrides this.

ChrisKnightLDN:
Agree - Toast and Alert. Two different components(?)

- As feedback that things are underway when the user initiates an asynchronous or background process. This will often be accompanied by a follow up message when the process completes. **Toast**
- Show minor warnings and errors when a background process encounters difficulty **Alert**
- To give end users hints on how to use a feature they have just activated. **Toast *but not sure about this use-case**
- To let end users know their slow or asynchronous action was successful **Toast**

PR #117, opened 2023-08-29
Edits to changelog

jamesscottbrown:
This adds keywords for change type, uses the actual component names for better searchability, and formats the component names.

Some of these `minor` changes should actually have been `major` (e.g., removing `TabSelector`), but the release has a version number now so too late to change.


PaulioRandall (approved):
LGTM

PR #118, opened 2023-08-29
Add square IconButton component

PaulioRandall:
**What does this change?**

Adds a square icon button.

**Why?**

Because the standard button is too cumbersome to overlay on maps and other visualizations.

**Use cases**

- Primarily to overlay on visualizations.
  - As map control buttons, e.g. panning, zoom, fullscreen, etc.
  - Possible on the edge of graphs to toggle modes, e.g. switch between rolling average and accumulative or between discrete value and percentage.
- Can be used on normal pages where space is a premium or only icon labels are needed. The standard button component should be considered first however.

**How?**

Just a ldn-viz styled square button.

Button size is clamped so the button is not too small or too big when users have tiny or huge font-size set in their browser preferences.

**Related issues**:

None.

None.

**How is it tested?**

Stories.

**How is it documented?**

Stories.

Not yet.

**Is it complete?**
- [ x] If this adds a new component, is it exported via `index.js`?


jamesscottbrown:
The `base` size isn't quite the same as the `base` size of the main `Button` component, but perhaps this doens't matter as they're intended to be used in different contexts where they'll be spatially separated.

jamesscottbrown:
This has now been superceeded by the redesigned`Button` component.

PR #121, opened 2023-08-30
Delete packages/ui/.svelte-kit directory

jamesscottbrown:
I initially commited this directory in an attempt to fix the GitHub Action that builds the Storybook site and deploys it to Github pages.

However, I think this was unnecessary: I forked this repo into my account's namespace, deleted the directory from the fork, and confirmed that the deployment still works.

This directory doesn't need to be kept under version-control as it gets automatically created by SvelteKit, and including it results in a lot of uncessary distractions being added to PRs.

If we merge this PR into `dev`, we can then merge `dev` into the feature branches with open PRs, so that they do not include modifications to these files.


PaulioRandall:
Ahh. I noticed this and deleted it in one of my branches only to find `npm run docs` fails due to missing type stuff. So I put it back and left it.

Will have to remember to build before running storybooks. Might want to add something to the README if it's not already in there.

PaulioRandall (approved):

ChrisKnightLDN:
Agree this should not be being committed.

PR #126, opened 2023-10-04
in CheckboxGroup replace select/clear all buttons with checkbox

jamesscottbrown:
In `CheckboxGroup` component , replace select/clear all buttons with a checkbox.

This brings the implementation closer to the design in Figma.

**Is it complete?**


jamesscottbrown:
Perhaps the `buttonsHidden` should now be renamed?

Issue #132, opened 2023-10-06
Improve display of button stories in storybook


ChrisKnightLDN:
Buttons have many variations and the stories kind of munge them together. Maybe a table layout or matrix or... would be better

PR #141, opened 2023-10-09
Table of buttons

jamesscottbrown:
This adds a story showing off the new button designs, similar to the table in Figma - related to #132.

Screenshot 2023-10-09 at 15-10-16 Ui _ Button - Sumamry table ⋅ Storybook

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,19 +60,6 @@
         "vite": "^4.2.0"
       }
     },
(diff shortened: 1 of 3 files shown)


ChrisKnightLDN:
I updated styles in the story - but more importantly I though that the Types should be respected so I separated the handling of 'disabled' state from condition prop.

PR #148, opened 2023-10-16
add NonIdealState component to act as a placeholder that explains why content cannot be displayed

jamesscottbrown:
**What does this change?**

Adds a NonIdealState component to act as a placeholder that explains why content cannot be displayed.

**Why?**

In some apps (e.g., the new multivariate census explorer) it would be useful to show placeholders which explain that user action is required before visualizations can be displayed.

Also, for some apps it would be useful to display something if there are unrecoverable data-fetching errors that would block everything.

No.

**How is it tested?**

**How is it documented?**

With Storybook stories.

**Is it complete?**


jamesscottbrown:
See issue #150

jamesscottbrown:
See preview at: https://dev.ldn-gis.co.uk/storybook-non-ideal-state/?path=/docs/ui-nonidealstate--documentation

PR #149, opened 2023-10-16
Codespaces config

jamesscottbrown:
This adds a config file so that starting a Codespace for this repo installs the required VSCode extensions.


jamesscottbrown:
I'm closing this PR: these changes slow down buillding a dev container (unless these are cached as "prebuilds", but for repos within an organization this [requires a paid plan](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds)).

PR #153, opened 2023-10-23
Reformat files with prettier

jamesscottbrown:
@PaulioRandall pointed out that while prettier doesn't report any warnings or errors for these files (and says "All matched files use Prettier code style!") , `prettier --write` does make changes.


jamesscottbrown:
I think @ChrisKnightLDN and I didn't notice this because our workflow is to automatically run prettier to format the files we've worked on when they are saved (and then to check formatting with prettier before making a PR) rather to run `prettier --write` or `npm run format` to reformat everything.

PaulioRandall (approved):
LGTM

PR #156, opened 2023-10-23
Button titles

PaulioRandall:
**What does this change?**

Adds `title` property to `Button` component so that simple lightweight and accessible text tooltips can be applied to buttons.

**Why?**

Because it may not be clear to all what a button with only an icon symbolizes. A heavy weight tooltip should be avoided for simple use cases, e.g. from libraries such as _Tippy.js_ or _PopperJS_.

**How?**

Uses the inbuilt `title` property and HTML specification rules.

**Related issues**:

None.

No.

**How is it tested?**

Storybook.

**How is it documented?**

Added to an existing Storybook story.

N/A.

**Is it complete?**
- [ x] Have you included changeset file?


jamesscottbrown (approved):
This looks fine.: the ability to set the `title` might sometimes be useful.

We generally favour directly labelling buttons with words (rather than using icons alone), but square buttons might be an exception

PR #170, opened 2023-11-28
Add Maplibre Map component

PaulioRandall:
**What does this change?**

Adds a lightweight `<Map>` component for creating MapLibre maps.

**Why?**

So I can build maps.

**How?**

Two new components:
- `Map` that simplifies the creation an initialization of a MapLibre map using Svelte.
- `MapApp` that wraps `Map` and provides page layout for full page maps.

**Related issues**:

None.

None.

**How is it tested?**

In storybook.

**How is it documented?**

In storybook.

Nope.

**Is it complete?**

> This question is redundant. It's complete when we agree it's complete.

- [x ] Have you included changeset file?


packages/maps/src/lib/themes/bounds.js
```diff
@@ -1,16 +1,53 @@
...
+
+// scaleBounds returns a new bounds that scales the input bounds up by the
+// specified amount { x, y } with the center as the origin.
+//
+// Given the amounts:
+// - { x: 0.1, y: 0.1 } then both width and height will be 10% bigger;
+// - { x: 0.5, y: 1 } then width will be 50% bigger and height 100% bigger;
+// - { x: 2, y: 0 } then width will be 200% bigger and height left unchanged;
+// - etc.
+export const scaleBounds = (bounds, amount = { x: 0, y: 0 }) => {
```
jamesscottbrown, line 14: I'd expect a `scale` function to take scale factors, rather than `(1 - scale-factor)` (this is, for example, what the CSS `scale()` function accepts).

packages/maps/src/lib/themes/bounds.js
```diff
@@ -1,16 +1,53 @@
...
+export const GREATER_LONDON_ZOOM = 11;
 export const GREATER_LONDON_BOUNDS = [
-	[-0.489, 51.28],
-	[0.236, 51.686]
+	[-0.505, 51.285],
+	[0.33, 51.693]
 ];
-
-// These coordinates are in the form needed by the option to MapboxGeocoder
-export const GREATER_LONDON_BOUNDS_FLAT = [-0.489, 51.28, 0.236, 51.686];
```
jamesscottbrown, line 10: Why is this deleted?

jamesscottbrown:
The `Map` component takes an object of options as a prop

Since the `style` attribute must be specified for anything to be rendered, it is requird rather than optional and I don't think it belongs in an `options` object: either it should be a separate prop, or a default value should be used if `options.style` is undefined.

packages/maps/src/lib/themes/bounds.js
```diff
@@ -1,16 +1,53 @@
...
+		]
+	];
+};
+
+export const GREATER_LONDON_ZOOM = 11;
 export const GREATER_LONDON_BOUNDS = [
-	[-0.489, 51.28],
-	[0.236, 51.686]
+	[-0.505, 51.285],
+	[0.33, 51.693]
```
jamesscottbrown, line 38:
```suggestion
	[-0.517, 51.278],
	[0.342, 51.698]
```


Where do these new values come from?

Running one of these SQL queries:

```sql
select st_transform(ST_Envelope(shape), 4326) from gisapdata.core_gla_boundary;
select st_transform(ST_SetSRID(ST_extent(geom), 27700), 4326) from gisapdata.core_london_borough;
```

indicates that the corners of a bounding-box would be:
```
-0.5160678730533503
 0.3413716859178191

51.27878314684388
51.69760660683185
```
Rounding to 3 d.p., by rounding the lower-bounds down and the upper-bounds up (i.e. both away from 0) gives:

```
-0.517
 0.342

51.278
51.698
```

The currently propsoed bound sclip part of Havering and Hillingdon:

```sql
select * from gisapdata.core_london_borough where not st_coveredBy(geom, st_transform(ST_MakeEnvelope(-0.505, 51.285, 0.33, 51.693, 4326), 27700));
-- returns Havering and Hillingdon
```

Whereas the bounds suggested in this comment don't:

```sql
select * from gisapdata.core_london_borough where not st_coveredBy(geom, st_transform(ST_MakeEnvelope(-0.517, 51.278, 0.342, 51.698, 4326), 27700));
-- no results

```

packages/maps/src/lib/themes/bounds.js
PaulioRandall: You can use `GREATER_LONDON_BOUNDS.flat()` (like the default options in the `<Map>` component).

packages/maps/src/lib/themes/bounds.js
PaulioRandall: My intention was to include all of Greater London based on the borough boundary context layer since that is what users will see when the map loads. The original coords cut off a chunk of Havering.

It was a quick eyeball test in mobile view. **You're welcome to generate a more precise set of coords if you like.**

packages/maps/src/lib/map/Map.svelte
```diff
@@ -1 +1,87 @@
...
+<script context="module">
+	export const appendOSKeyToUrl = (osKey) => {
+		return (url) => {
+			url = new URL(url);
+
+			if (!url.hostname === 'api.os.uk') {
+				return { url: url.href };
+			}
+
+			const WGS84 = 3857;
```
jamesscottbrown, line 10: This line is misleading.

`EPSG 3857` is the projected coordinate system commonly referred to as Web Mercator, whereas I think most people would expect `WGS84` to refer to `EPSG 4326`.

These are fundamentally different things: 4326 uses angular units (degrees) to represent positions on an ellipsoid; 3857 uses linear units (m) to represent positions projected onto a plane.

packages/maps/src/lib/map/Map.svelte
```diff
@@ -1 +1,87 @@
...
+			url.searchParams.set('srs', WGS84);
```
jamesscottbrown, line 11:
```suggestion
			url.searchParams.set('srs', 3857); // Web Mercator
```

packages/maps/src/lib/themes/bounds.js
PaulioRandall: Correction, the `<Map>` component isn't using flat bounds anymore.

packages/maps/src/lib/map/Map.svelte
PaulioRandall: I see, I've updated the naming.

packages/maps/src/lib/map/Map.svelte
PaulioRandall: See comment above.

packages/maps/src/lib/themes/bounds.js
jamesscottbrown: > **You're welcome to generate a more precise set of coords if you like**

I have - they're in my Suggested Change above. I'll click the `Commit Suggestion` button on it.

packages/maps/src/lib/themes/bounds.js
PaulioRandall: I've updated the API & implementation to use scale factors. I also added some simple tests.

PaulioRandall:
> The `Map` component takes an object of options as a prop.
> 
> Since the `style` attribute must be specified for anything to be rendered, it is required rather than optional and I don't think it belongs in an `options` object: either it should be a separate prop, or a default value should be used if `options.style` is undefined.

While you're technically correct on the required/optional note, I crafted the `<Map>` component API so it's consistent with the `maplibre_gl.Map` class constructor API. I think this is the most readable and flexible approach. I 'm trying to deviate as little as possible to avoid confusion. It also allows us to reuse [MapLibre documentation](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.MapOptions/) with the only exception being the `container` prop being set internally.

jamesscottbrown:
> While you're technically correct on the required/optional note, I crafted the `<Map>` component API so it's consistent with the `maplibre_gl.Map` class constructor API. I think this is the most readable and flexible approach. I 'm trying to deviate as little as possible to avoid confusion. It also allows us to reuse [MapLibre documentation](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.MapOptions/) with the only exception being the `container` prop being set internally.

That is reasonable, but could we use `os_light_vts` as the default if `style` isn't specified?

PR #172, opened 2023-12-01
Map controls

PaulioRandall:
**What does this change?**

Adds map controls including:
- `MapControlFullscreen`
- `MapControlPan`
- `MapControlRefresh`
- `MapControlZoom`

Also adds `MapControlGroup` that allows for sets of map controls to be grouped and positioned within a map.

**Why?**

Clickable controls for the map.

**How?**

I don't know.

**Related issues**:

None.

No.

**How is it tested?**

Storybook + within an application.

**How is it documented?**

Storybook.

Yes but it's complicated.

- [x ] Have you included changeset file?
- [x ] If this adds a new component, is it exported via `index.js`?


jamesscottbrown:
I think the uses of `on:keypress={newKeyHandler(zoomIn)}` are unnecessary: the button component will fire an `onClick` event when the user presses the Enter or Space keys while it has focus.

I agree with @mikeldn that for the pan buttons a three-row/diamond-shaped arrangement would be better than the current two-row/triangle-shaped arrangement.

I'm not entirely sure about the behaviour of the fullscreen button. It might be better to have a separate buttons for opening in full-screen and opening in the parent of the iframe (which could be auotmatically hidden if not in an iframe): as a user, I'd generally prefer to just full-screen the map, though there are use-cases for both.

PaulioRandall:
We don't do enough on the accessibility front. I want to add more such enhancements but the current interface to the `Button` component prevents me due to lack of `{...$$restProps}`.

jamesscottbrown:
This looks fine.

PR #182, opened 2024-01-10
Change headlessui library

jamesscottbrown:
**What does this change?**

**Why?**

**How?**

**Related issues**:

**How is it tested?**

**How is it documented?**

**Is it complete?**

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -60,7 +60,6 @@
 	"main": "./dist/index.js",
 	"type": "module",
 	"dependencies": {
-		"@rgossiaux/svelte-headlessui": "^1.0.2",
 		"@steeze-ui/heroicons": "^2.2.3",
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
Some thoughts:

* like the example, this uses `svelte-transition` to transition the opacity of the modal between 0 and 100. We haven't yet given much thought to either what animated transitions should be used in our design system, or whether we should use a library like `svelte-transition` to implement them
* this changes the API of the modal component, from exposing an `isOpen` Boolean, to exposing an object with `.open()` and `.close()` methods
* is this the best headless UI library for us to use, or should we consider using [Melt UI](https://melt-ui.com) instead?

ChrisKnightLDN:
Melt-ui looks interesting I'd like to try it out

ChrisKnightLDN:
Closing in favour of pursuing melt-ui

PR #184, opened 2024-01-11
initial re-implementation of the LDN-squared borough tile map from viz-tools

jamesscottbrown:
**What does this change?**

**Why?**

**How?**

**Related issues**:

**How is it tested?**

**How is it documented?**

**Is it complete?**


ChrisKnightLDN:
this is known colloquially as LDN-squared to avoid confusion we should probably call it that where needed

jamesscottbrown:
I think https://github.com/Greater-London-Authority/ldn-viz-tools/pull/565 is probably a better approach.

PR #185, opened 2024-01-12
add color scheme code from observable notebook

jamesscottbrown:
This adds the definitions of  `ldnColors`, `theme`, `getColorScale()`, and `getThresholdBreaksColorsLabels()` from https://observablehq.com/@ldn/color-system, so that they can be easily imported by chart components or projects.

After this has been published to npm, the notebook can be updated to import the functions form `@ldn-viz/charts`, so that there is a single source-of-truth.

The functions `swatch()`/`swatches()`/`swatchGrid()`/`asRamp()`/`ramp()` haven't been added, as they are specific to showing color schemes/ramps in Observable notebooks.

### DIFF
--- a/.changeset/tidy-lizards-leave.md
+++ b/.changeset/tidy-lizards-leave.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/charts': minor
+---
+
+ADDED - color theme object and functions for generating color scales
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,16 +62,134 @@
         "vite": "^4.2.0"
       }
     },
+    "apps/docs/node_modules/@sveltejs/kit": {
+      "version": "1.30.1",
+      "dev": true,
+      "hasInstallScript": true,
+      "license": "MIT",
+      "dependencies": {
+        "@sveltejs/vite-plugin-svelte": "^2.5.0",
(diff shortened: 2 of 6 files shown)


ChrisKnightLDN:
Wonder if this isn't better facillitated in themes? (where colors are already defined) then all the functions would be universally available to packages from a central location..?

jamesscottbrown:
The color scales probably belond here, as they're mappings from data to colors, so is inherently visualization focused.

The theme definitions could potentially be moved, but they seem quite vis-focused (e.g., they have `axis`/`grid`/`labels` attributes).

ChrisKnightLDN:
The theme definitions would ultimately be 'themes' expanded upon with other tokens. Those here are currently chart focussed, but eventually they should be centralised along with all the other tokens - which will be centrally in 'themes'. So they could start here, but they'd have to be moved. Timeline for all this tbd.

ChrisKnightLDN:
(also - colour scales valuable for maps?) And other potential ui elements that are perhaps not full charts - these might live in ui. so the colorscales could be consumed by all three packages - making 'themes' the best candidate for their location imho. unless we have some other utilities global area - but I don't think we should over complicate.

jamesscottbrown:
We've agreed to move this to a separate `utils` package.

PR #186, opened 2024-01-15
move chrome that goes around chart from shared/ to chartContainer/

jamesscottbrown:
**What does this change?**

The `charts` library includes a `chartContainer` component that adds chrome like a Title/SubTitle/Export buttons around a chart.

This PR moves the components used by this component from the `shared/` directory to the `chartContainer` directory.


ChrisKnightLDN:
What's the logic of this as the chart container is 'shared'? (asking for a friend)

jamesscottbrown:
They are all used to render things *outside* the main chart area, and are currently used only by the `chartContainer`, so it makes sense to group them together.

In contrast, the rest of the contents of `shared/` are to be used *inside* the plot area, within a `LayerCake` component.

PR #189, opened 2024-01-15
create utils package including color functions

jamesscottbrown:
**What does this change?**

This adds the definitions of  `ldnColors`, `theme`, `getColorScale()`, and `getThresholdBreaksColorsLabels()` from https://observablehq.com/@ldn/color-system, so that they can be easily imported by chart components or projects.

After this has been published to npm, the notebook can be updated to import the functions form `@ldn-viz/charts`, so that there is a single source-of-truth.

The functions `swatch()`/`swatches()`/`swatchGrid()`/`asRamp()`/`ramp()` haven't been added, as they are specific to showing color schemes/ramps in Observable notebooks.

This replaces https://github.com/Greater-London-Authority/ldn-viz-tools/pull/185.


jamesscottbrown:
We've decided to move these color functions to the `themes` package instead.

jamesscottbrown:
Actually we'll put them in utils.

mikeldn:
@jamesscottbrown hmm... good point. It's not really returning a scale is it.. you're right. I think getColorRamp() is best. If we used scheme, you might expect to be able to pass it in ('blue') and get back a blue scheme... which isn't a terrible idea actually :) (But actually that is in ldnColors.perceptual.blue.... (all of which need work).

so, yes, getColorRamp() is good. Don't forget that getThresholdBreaksColorsLabels() uses it, so will need changing too.

PR #190, opened 2024-01-15
make width of obsservable plots optionally responsive

jamesscottbrown:
**What does this change?**

Provides a prop that allows the observable plot component's width to be repsonsive.

It debounces increases in width, but we can't debounce decreases, as observable plot will detect if the plot is wider than its parent and resize itself.


jamesscottbrown:
See preview: https://dev.ldn-gis.co.uk/storybook-observable-plot-responsive/?path=/story/charts-observableplot--with-responsive-width

PR #191, opened 2024-01-16
don't auto-remove imports on save in VSCode

jamesscottbrown:
**What does this change?**

After this change, VScode will still auto-sort imports on save, but won't automatically remove any unused imports.

**Why?**

Automatically removing imports can be quite annoying - for example, when commenting out a sction of code duing debugging.


ChrisKnightLDN:
As annoying as this is, I would advocate that we keep it. Having unused imports in files, or manually having to groom imports is more annoying imo. Or at least it's worth the pain.

jamesscottbrown:
I think eslint will still warn about the missing imports, to tell you to remove them before committing/merging.

Unused imports should be removed before commiting/merging, but removing them on every save seems unnecessary.

PR #201, opened 2024-01-19
Docs update

ChrisKnightLDN:
**What does this change?**
Adds an index route to each package.
Adds tailwind setup to maps.
Updates tests in all packages

**Why?**
Running npm run dev caused terminal errors. Adding stub index pages fixes (is consistent with sveltekit library starter)
There should be parity in tailwind setup between packages

- [X] Have you included changeset file?


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-docsUpdate/

PR #206, opened 2024-01-26
Fix MapApp & Map layouts

PaulioRandall:
**What does this change?**

1. Fixed _MapApp_ not able to accommodate full width headers and footers.
2. Fixed _Map_ not hiding overflow caused by popups and tooltips on the map.

Also removed `id` from _Map_ as to accommodate multiple maps per page.

**Why?**

It broken.

### DIFF
--- a/.changeset/abc-efg-hij.md
+++ b/.changeset/abc-efg-hij.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
LGTM. Preview: https://dev.ldn-gis.co.uk/storybook-fix-mapapp-layout/?path=/docs/maps-map--documentation

PR #207, opened 2024-01-26
Update Checkbox so name prop required and id is optional

### LINKED ISSUE #205: Checkbox.svelte is missing 'name' field and 'id' should be optional
**Describe the issue**

Regarding `Checkbox.svelte`:

1. Required _name_ field is missing. Can't submit a form using standard HTML form mechanics without it.
2. _id_ field should be optional as not always needed.

PaulioRandall:
**What does this change?**

Regarding _Checkbox_ component:

1. Adds a required `name` prop.
2. Promotes `name` prop to primary identifier.
3. Demotes `id` prop to optional.

**Why?**

HTML form fields must have a `name` attribute as it is paired with the field value during submission, e.g. `name1=value1&name2=value2`. `name` must be unique within its `<form>`.

The `id` is not used by the form but still useful for general element stuff, e.g. CSS & JS selector.

**Related issues**:

Resolves #205.

**How is it tested?**

Storybook.

**How is it documented?**

Storybook.

### DIFF
--- a/.changeset/jjj-kkk-lll.md
+++ b/.changeset/jjj-kkk-lll.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': patch
+---
+
+Add 'name' attribute to Checkbox and set 'id' as optional.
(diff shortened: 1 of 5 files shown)


PaulioRandall:
@jamesscottbrown This still needs to be done but I think this branch is too out of date. It's a small change so I will do it in a new branch and PR.

PR #219, opened 2024-03-01
Fix map controls divs interferring with map interactions

PaulioRandall:
**What does this change?**

Fixes map controls interfering with map interactions.

**Why?**

Map control areas preventing panning and drawing on the map. Mostly only noticable on smaller laptops, e.g. surface pro etc.

**How?**

Disabled all pointer events in map controls and enable only at a fine grained level, i.e. applied pointer events to buttons directly rather than their containing elements to avoid padded and margin space from causing problems too.

### DIFF
--- a/packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
+++ b/packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
@@ -29,7 +29,7 @@
 </script>
 
 <script lang="ts">
-	export let position: string | MapControlGroupPositions = MapControlGroupPositions.TopLeft;
+	export let position: keyof typeof MapControlGroupPositions = MapControlGroupPositions.TopLeft;
 	export let classes = '';
 
 	const positionClass = positionClasses[position];
--- a/packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
+++ b/packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
@@ -12,7 +12,7 @@
 	}
 
 	type PositionClass = {
-		[key: string]: string;
+		[key in keyof typeof MapControlGroupPositions]: string;
 	};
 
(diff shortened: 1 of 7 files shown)


packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
```diff
@@ -12,7 +12,7 @@
 	}
 
 	type PositionClass = {
-		[key: MapControlGroupPositions]: string;
+		[key: string]: string;
```
jamesscottbrown, line 15:
```suggestion
		[key in keyof typeof MapControlGroupPositions]: string;
```

This has the effect of defining`PositionClass` to be:

```ts
type PositionClass = {
    readonly TopLeft: string;
    readonly TopCenter: string;
    readonly TopRight: string;
    readonly TopRightOffset: string;
    readonly CenterRight: string;
    readonly BottomRight: string;
    readonly BottomCenter: string;
    readonly BottomLeft: string;
    readonly CenterLeft: string;
}
```

packages/maps/src/lib/mapControlGroup/MapControlGroup.svelte
```diff
@@ -29,12 +29,12 @@
 </script>
 
 <script lang="ts">
-	export let position = MapControlGroupPositions.TopLeft;
+	export let position: string | MapControlGroupPositions = MapControlGroupPositions.TopLeft;
```
jamesscottbrown, line 32:
```suggestion
	export let position: keyof typeof MapControlGroupPositions = MapControlGroupPositions.TopLeft;
```

jamesscottbrown:
This looks fine.

The linting Github Action seems to have got stuck: there is apparently a problem with Actions currently.

image
https://www.githubstatus.com/

jamesscottbrown:
LGTM

PR #224, opened 2024-03-05
Location search

PaulioRandall:
> TODO

**What does this change?**

**Why?**

**How?**

**Related issues**:

**How is it tested?**

**How is it documented?**

**Is it complete?**

### DIFF
--- a/packages/maps/src/lib/mapControlLocationSearch/Geocoder.svelte
+++ b/packages/maps/src/lib/mapControlLocationSearch/Geocoder.svelte
@@ -1,6 +1,8 @@
-<script context="module">
-	const newDebouncer = (func, delay) => {
-		let timerId = null;
+<script lang="ts" context="module">
+	type Timeout = ReturnType<typeof setTimeout>;
+
+	const newDebouncer = (func: () => void, delay: number) => {
+		let timerId: undefined | Timeout = undefined;
 
 		const restart = () => {
 			clearTimeout(timerId);
@@ -15,12 +17,12 @@
 	};
 </script>
 
-<script lang='ts'>
+<script lang="ts">
 	import { MagnifyingGlass } from '@steeze-ui/heroicons';
 	import { Icon } from '@steeze-ui/svelte-icon';
 	import GeocoderSuggestionList from './GeocoderSuggestionList.svelte';
-		import type { GeocoderAdapter } from './GeocoderAdapter'
-			import type { MapControlLocationUpdate } from './types'
+	import type { Geolocation, GeocoderAdapter, GeolocationNamed } from './GeocoderAdapter';
+	import type { OnSearchResult } from './types';
 
 	export let adapter: GeocoderAdapter;
 
@@ -29,11 +31,11 @@
 
 	// onLocationSelected is called when a user clicks a location from the
 	// suggestions list.
-	export let onLocationSelected: MapControlLocationUpdate = undefined;
+	export let onLocationSelected: OnSearchResult = undefined;
 
 	// results can be bound via 'bind:results' to be notified of new search
 	// results reactively.
-	export let results = null;
+	export let results: null | GeolocationNamed[] = null;
 
 	// disableSuggestionList can be used to disable to dropdown suggestion
 	// list. The suggestion property will still be updated.
@@ -64,7 +66,7 @@
 		try {
 			results = (await adapter.search(query)) || [];
 			showSuggestionList = !disableSuggestionList;
-		} catch (e: Error) {
+		} catch (e: unknown) {
 			console.error(e);
 			showSuggestionList = false;
 		}
@@ -77,7 +79,7 @@
 		showSuggestionList = false;
 	};
 
-	const onSelect = (suggestion) => {
+	const onSelect = (suggestion: Geolocation) => {
 		closeSuggestionsList();
 
 		if (onLocationSelected) {
@@ -85,14 +87,15 @@
 		}
 	};
 
-	const hideSuggestionList = (event) => {
-		if (!isElementOrChildOf(event.target, container)) {
+	const hideSuggestionList = (event: MouseEvent | KeyboardEvent) => {
+		const target = event.target as HTMLElement;
+		if (!target || !container || !isElementOrChildOf(target, container)) {
 			closeSuggestionsList();
 		}
 	};
 
-	const isElementOrChildOf = (element, possibleParent) => {
-		const position = element.compareDocumentPosition(possibleParent);
+	const isElementOrChildOf = (element: HTMLElement, parent: HTMLElement) => {
+		const position = element.compareDocumentPosition(parent);
 		return !!(position & Node.DOCUMENT_POSITION_CONTAINS);
 	};
 
@@ -102,13 +105,16 @@
 		}
 	};
 
-	const selectFirstSuggestion = (event) => {
+	const selectFirstSuggestion = (event: KeyboardEvent) => {
 		if (!disableSuggestionList && results && results.length > 0 && event.key === 'Enter') {
 			onSelect(results[0]);
 		}
 	};
 
-	$: scheduleUpdate(query);
+	$: {
+		query;
+		scheduleUpdate();
+	}
 </script>
 
 <svelte:window on:mousedown={hideSuggestionList} on:keyup={hideSuggestionList} />
--- a/packages/maps/src/lib/mapControlLocationSearch/Geocoder.svelte
+++ b/packages/maps/src/lib/mapControlLocationSearch/Geocoder.svelte
@@ -1,0 +1,138 @@
+<script context="module">
+	const newDebouncer = (func, delay) => {
+		let timerId = null;
+
+		const restart = () => {
+			clearTimeout(timerId);
+			timerId = setTimeout(func, delay);
+		};
+
+		const clear = () => {
+			clearTimeout(timerId);
+		};
+
+		return [restart, clear];
+	};
+</script>
+
+<script lang='ts'>
+	import { MagnifyingGlass } from '@steeze-ui/heroicons';
+	import { Icon } from '@steeze-ui/svelte-icon';
+	import GeocoderSuggestionList from './GeocoderSuggestionList.svelte';
+		import type { GeocoderAdapter } from './GeocoderAdapter'
+			import type { MapControlLocationUpdate } from './types'
+
+	export let adapter: GeocoderAdapter;
+
+	// delay in ms after a key stroke to minimise redundant API requests.
+	export let delay = 250;
+
+	// onLocationSelected is called when a user clicks a location from the
+	// suggestions list.
+	export let onLocationSelected: MapControlLocationUpdate = undefined;
+
+	// results can be bound via 'bind:results' to be notified of new search
+	// results reactively.
+	export let results = null;
+
+	// disableSuggestionList can be used to disable to dropdown suggestion
+	// list. The suggestion property will still be updated.
+	//
+	// This is useful if you're binding to the results property and showing
+	// the results in custom way.
+	export let disableSuggestionList = false;
+
+	// classes is a space separated list of additional classes applied to the
+	// root container.
+	export let classes = '';
+
+	// inputClasses is a space separated list of additional classes applied to
+	// the input element because laying input elements can be quite fiddly.
+	export let inputClasses = '';
+
+	let container: HTMLElement;
+	let query = '';
+	let showSuggestionList = false;
+
+	const updateSuggestionsNow = async () => {
+		if (!query || query.length < 3) {
+			results = null;
+			showSuggestionList = false;
+			return;
+		}
+
+		try {
+			results = (await adapter.search(query)) || [];
+			showSuggestionList = !disableSuggestionList;
+		} catch (e: Error) {
+			console.error(e);
+			showSuggestionList = false;
+		}
+	};
+
+	const [scheduleUpdate, cancelUpdate] = newDebouncer(updateSuggestionsNow, delay);
+
+	const closeSuggestionsList = () => {
+		cancelUpdate();
+		showSuggestionList = false;
+	};
+
+	const onSelect = (suggestion) => {
+		closeSuggestionsList();
+
+		if (onLocationSelected) {
+			onLocationSelected(suggestion);
+		}
+	};
+
+	const hideSuggestionList = (event) => {
+		if (!isElementOrChildOf(event.target, container)) {
+			closeSuggestionsList();
+		}
+	};
+
+	const isElementOrChildOf = (element, possibleParent) => {
+		const position = element.compareDocumentPosition(possibleParent);
+		return !!(position & Node.DOCUMENT_POSITION_CONTAINS);
+	};
+
+	const reshowSuggestionList = () => {
+		if (results) {
+			showSuggestionList = !disableSuggestionList;
+		}
+	};
+
+	const selectFirstSuggestion = (event) => {
+		if (!disableSuggestionList && results && results.length > 0 && event.key === 'Enter') {
+			onSelect(results[0]);
+		}
+	};
+
+	$: scheduleUpdate(query);
+</script>
+
+<svelte:window on:mousedown={hideSuggestionList} on:keyup={hideSuggestionList} />
+
+<search
+	bind:this={container}
+	class="bg-core-grey-600 pointer-events-auto relative w-full h-10 flex {classes}"
+>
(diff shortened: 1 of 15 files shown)


jamesscottbrown:
Previews: https://dev.ldn-gis.co.uk/storybook-location-search/?path=/docs/maps-mapcontrollocationsearch--documentation

I think there should be prop that determines whether the "Find my location" button is displayed, as it isn't always wanted. It is most useful for apps aimed at the genral public who are interested in things near them (e.g., the cool-spaces map); it's typically less useful for applications used by council officers who are interested in some specific location that is typically not where they currently are.


There seem to be a couple of bugs affecting keyboard interactions:

* after typing in the input box, I should be able to use the down arrow keys to select the first item, and then use up/down to change which item is selected (as I can in the original Mapbox component); currently this doens't work

*  once an item has been selected, pressing the enter key should both navigate to that location and close the drop-down list of location results (as happens in the original MapBox component, or when the user makes a choice by clicking rather than pressing enter); currently it navigates to the location, but leaves the drop-down list open


I'll have a look at the code later.

.changeset/aaa_bbb_ccc_ddd.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ldn-viz/ui': minor
+---
```
jamesscottbrown, line 3:
```suggestion
'@ldn-viz/maps': minor
'@ldn-viz/ui': minor
---
```

This PR updates both `maps` and `ui`

packages/maps/src/lib/mapControlLocationSearch/map-types.ts
```diff
@@ -0,0 +1,14 @@
+import type { Writable } from 'svelte/store';
+
+import type {
+	Map, //
```
jamesscottbrown, line 4:
```suggestion
	Map,
```

PaulioRandall:
> Previews: https://dev.ldn-gis.co.uk/storybook-location-search/?path=/docs/maps-mapcontrollocationsearch--documentation
> 
> I think there should be prop that determines whether the "Find my location" button is displayed, as it isn't always wanted. It is most useful for apps aimed at the general public who are interested in things near them (e.g., the cool-spaces map); it's typically less useful for applications used by council officers who are interested in some specific location that is typically not where they currently are.
> 
> There seem to be a couple of bugs affecting keyboard interactions:
> 
> * after typing in the input box, I should be able to use the down arrow keys to select the first item, and then use up/down to change which item is selected (as I can in the original Mapbox component); currently this doens't work
> * once an item has been selected, pressing the enter key should both navigate to that location and close the drop-down list of location results (as happens in the original MapBox component, or when the user makes a choice by clicking rather than pressing enter); currently it navigates to the location, but leaves the drop-down list open
> 
> I'll have a look at the code later.

1. It currently uses the tab and shift+tab. Extending to use the up and down arrows shouldn't be a problem.
2. By selected, do you mean hovered?
  - It's subtle but both MapBox and this component have two separate highlights: hover and focus (they use the same colour)
  - The current behavior when clicking _'enter'_ while input is focused takes the user to the first location.

I'll apply the arrow keys then we'll have another play.

packages/maps/src/lib/mapControlLocationSearch/map-types.ts
jamesscottbrown: after removing the empty comment, Prettier might want this import statement on a single line

packages/ui/src/lib/geolocation/types.ts
```diff
@@ -0,0 +1,44 @@
+// Coords represents map coordinates in the form [lng, lat].
+export type GeolocationCoords = [number, number];
```
jamesscottbrown, line 2:
```suggestion
export type GeolocationCoords = [lng: number, lat: number];
```

packages/ui/src/lib/geolocation/types.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Boudns represents map bounds in the form [minLng, minLat, maxLng, maxLat].
+export type GeolocationBounds = [number, number, number, number];
```
jamesscottbrown, line 5:
```suggestion
export type GeolocationBounds = [minLng: number, minLat: number, maxLng: number, maxLat: number];
```

packages/ui/src/lib/geolocation/types.ts
```diff
@@ -0,0 +1,44 @@
...
+	center: GeolocationCoords;
+	bounds?: GeolocationBounds;
+}
+
+// GeolocationNamed represents a named and possible addressed location. Named
+// locations must be unique within a collection.
+export interface GeolocationNamed extends GeolocationUnamed {
+	id: string;
+
+	// name is the short human readable name of the loaction presented to the
```
jamesscottbrown, line 19:
```suggestion
	// name is the short human-readable name of the location presented to the
```

packages/ui/src/lib/geolocation/types.ts
```diff
@@ -0,0 +1,44 @@
...
+	// user.
+	name?: string;
+
+	// address in the human readable address presented to the user.
```
jamesscottbrown, line 23:
```suggestion
	// address in the human-readable address presented to the user.
```

packages/ui/src/lib/geolocation/types.ts
```diff
@@ -0,0 +1,44 @@
...
+// Geolocation represents either a named or unamed location.
+export type Geolocation = GeolocationUnamed | GeolocationNamed;
+
+// GeolocationSearchError represents an error thrown by a geolocation or
+// or geolocator search.
+export type GeolocationSearchError = Error | GeolocationPositionError;
+
+// OnGeolocationSearchResult is invoked when a location search feature
+// (Geocoder or Geolocator) finds a result or a suggested result is selected by
+// the user.
```
jamesscottbrown, line 39:
```suggestion
// OnGeolocationSearchResult is invoked when a Geocoder or Geolocator
// finds a result or a suggested result is selected by the user.
```

packages/maps/src/lib/mapControlLocationSearch/MapGeocoderAdapterMapBox.ts
```diff
@@ -0,0 +1,123 @@
...
+const transformGeoJSONToGeolocationNameds = (
+	geojson: MapBoxFeatureCollection
+): GeolocationNamed[] => {
+	return geojson.features.map((loc) => {
+		return {
+			id: loc.id,
+			name: loc.text,
+			address: removeNameFromAddress(loc.place_name, loc.text),
+			// loc.center isn't always the center of the bbox
+			center: calcCorrectCenter(loc.bbox, loc.center),
```
jamesscottbrown, line 94: This implies that the "correct" center of a feature is the center of the bounding box, which is not in general true.


If a feature's shape is not convex, then the centre of the bounding box might not even be inside the feature: consider a building shaped like the letter "n". 

I'm not sure whether or not adjusting the provided center positions like this is a good idea, but if it is then the function that does this should be called something like `calcBoundingBoxCenter` rather than `calculateCorrectCenter`.

packages/maps/src/lib/mapControlLocationSearch/map-layer.ts
```diff
@@ -0,0 +1,183 @@
+import { GLIDE_ANIMATION_OPTIONS } from '@ldn-viz/maps';
+import type {
+	GeolocationCoords, //
```
jamesscottbrown, line 3:
```suggestion
	GeolocationCoords,
```

packages/maps/src/lib/mapControlLocationSearch/map-layer.ts
```diff
@@ -0,0 +1,183 @@
...
+};
+
+const setMarker = (ref: string, map: Map, mapgl: MapGL, coords: GeolocationCoords) => {
+	clearMarker(ref);
+
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
+	markers[ref] = new mapgl.Marker() //
+		.setLngLat(coords)
+		.addTo(map); //
```
jamesscottbrown, line 113:
```suggestion
		.addTo(map);
```

packages/ui/src/lib/geolocation/GeocoderAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+// GeocoderAttribution specifies the attribution text and optional link
+// returned by an adapter.
+export interface GeocoderAttribution {
+	// text such as the company or individual name.
+	text: string;
+
+	// link URL to the data source. If this is unknown then link to the company
+	// or individual providing the service.
+	link?: string;
+}
```
jamesscottbrown, line 12:
```suggestion
export interface GeocoderAttribution {
	// attribution text such as the company or individual name.
	text: string;

	// optional link URL to the data source.
	// If this is unknown then link to the company or individual providing the service.
	link?: string;
}
```

packages/ui/src/lib/geolocation/GeocoderAdapterOSPlaces.ts
```diff
@@ -0,0 +1,240 @@
...
+import type { GeocoderAdapter } from './GeocoderAdapter';
+
+// LocalCustodianCode represents a mapping between a custodian code and its
+// readable label.
+export interface LocalCustodianCode {
+	label: string;
+	code: number;
+}
+
+// OS_LONDON_LOCAL_CUSTODIAN_CODES is an array of custodian codes mapped to
```
jamesscottbrown, line 11:
```suggestion
// OS_LONDON_LOCAL_CUSTODIAN_CODES is an array of AddressBase custodian codes mapped to
```

packages/ui/src/lib/geolocation/GeocoderAdapterOSPlaces.ts
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+const buildUrl = (text: string, key: string, lcc: number, resultCount: number): string => {
+	const queryString = new URLSearchParams({
+		query: encodeURIComponent(text),
+		key: key,
+		output_srs: 'WGS84',
+		format: 'JSON',
+		maxresults: resultCount.toString(),
```
jamesscottbrown, line 221: As for the MapBox adaptor, I think this use of `.toString()` is unncecessary

packages/ui/src/lib/geolocation/GeocoderSuggestion.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+	import type {
+		GeolocationNamed, //
+		OnGeolocationSearchResult
+	} from './types';
```
jamesscottbrown, line 5:
```suggestion
	import type { GeolocationNamed, OnGeolocationSearchResult } from './types';
```

packages/ui/src/lib/geolocation/Geolocator.svelte
```diff
@@ -0,0 +1,172 @@
+<script lang="ts">
+	import { onMount } from 'svelte';
+	import { Button, Spinner } from '@ldn-viz/ui';
+	import { XMark } from '@steeze-ui/heroicons';
+	import { Icon } from '@steeze-ui/svelte-icon';
+	import TargetIcon from './TargetIcon.svelte';
+
+	import type {
+		GeolocationCoords, //
```
jamesscottbrown, line 9:
```suggestion
		GeolocationCoords,
```

jamesscottbrown (commented):
* There is a method called `transformGeoJSONToGeolocationNameds`, which is quite awkward; `transformGeoJSONToNamedGeolocations` would be better.

* For the UI component, `LocationSearch` might be a more descriptive name than the current `Geocoder` (which captures the fact that it converts textual names to spatial corodinates, but not that it is UI element that prsents a search-box).

packages/maps/src/lib/mapControlLocationSearch/MapGeocoderAdapterMapBox.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const buildUrl = (text: string, token: string, resultCount: number): string => {
+	text = encodeURIComponent(text);
+
+	const queryString = new URLSearchParams({
+		access_token: token,
+		bbox: GREATER_LONDON_BOUNDS_PADDED.flat().toString(),
+		autocomplete: true.toString(),
+		limit: resultCount.toString()
```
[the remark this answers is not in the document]
PaulioRandall: Probably right, I'll change it and test.

packages/maps/src/lib/mapControlLocationSearch/MapGeocoderAdapterMapBox.ts
PaulioRandall: Hmmm, when searching for _'Canning Town'_ the symbol point is placed quite far away from the BBOX center which confused me (as a user). I know boundaries can be quite irregular.

I suppose it depends on the level of accuracy a user needs when location searching? As a user, you may want to pan around then come back to your searched spot. A centered symbol would be the better reference point.

In a future update we could put a border on the map representing the BBOX instead? Might look a bit ugly.

I'll rename the function to `calcBoundingBoxCenter`.

PaulioRandall: Or just not have a marker for BBOX locations?

packages/ui/src/lib/geolocation/GeocoderAdapterOSPlaces.ts
PaulioRandall: I'll remove and test

packages/maps/src/lib/mapControlLocationSearch/MapGeocoderAdapterMapBox.ts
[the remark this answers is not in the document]
PaulioRandall: Turns out this keeps TypeScript happy. If I try to use `as string` it gives lint errors.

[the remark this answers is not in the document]
jamesscottbrown: Ok - leave it as it is

packages/ui/src/lib/geolocation/Geocoder.svelte
```diff
@@ -0,0 +1,185 @@
...
+	const hideSuggestionList = (event: MouseEvent | KeyboardEvent) => {
+		const target = event.target as HTMLElement;
+		if (!target || !container || !isElementOrChildOf(target, container)) {
+			closeSuggestionsList();
+		}
+	};
+
+	const isElementOrChildOf = (element: HTMLElement, parent: HTMLElement) => {
+		const position = element.compareDocumentPosition(parent);
+		return !!(position & Node.DOCUMENT_POSITION_CONTAINS);
```
jamesscottbrown, line 114: Using `&` rather than `&&` here looks like a bug.

jamesscottbrown (commented):
Most of my comments have been addressed, but two are outstanding:

* I think `LocationSearch` accept a prop that determines whether the `Geolocator` is displayed, as it isn't always wanted (e.g., in applications used by council officers who are interested in some specific location that is typically not where they currently are).

* it should be possible to use the up/down arrow keys to move between suggestions, and between suggestions and the input field

packages/ui/src/lib/geolocation/Geocoder.svelte
PaulioRandall: Good spot. But the Bitwise AND is correct here: [https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition#example](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition#example)

PaulioRandall:
@jamesscottbrown I've updated so the arrow keys no function for moving up and down the suggestion list. It also allows extends to the input so ArrowUp will eventually take put you back into the end of the input element.

I had to refactor the event handling and messaging between the Geocoder, SuggestionList, and Suggestion components to make this happen.

PaulioRandall:
Added prop to hide Geolocator in maps too.

packages/ui/src/lib/geolocation/Geocoder.svelte
jamesscottbrown, line 202:
```suggestion
		const positionBitmask = element.compareDocumentPosition(parent);
		return !!(positionBitmask & Node.DOCUMENT_POSITION_CONTAINS);
```

Renaming to `positionBitmask` makes it clear that this is a bitmask, and hence that bitwise AND is intended to query the status of a specific bit.

PR #228, opened 2024-03-07
Add justValue prop to Select component

jamesscottbrown:
**What does this change?**

Add a `justValue` prop to the `Select` component that can be bound to.

The `value` prop is an object containing both the label and value of the selected item. This is inconvenient: 

Often the select sets a value that can also be updated by other interactions (for example, users might be able to select a borough by either using the select control or by clicking on a map). This is currently inconvenient to handle: we can't just bind a store to the `value` prop, since this is is an object containing both the label and value of the selected item, not just the value.

`svelte-select` provides a `justValue` prop, but it is read-only. This PR updates our wrapper around it (the `Select` component) to add a prop with the saem name that is writable.

**Is it complete?**

### DIFF
--- a/.changeset/hfuufh-ifie-eifie.md
+++ b/.changeset/hfuufh-ifie-eifie.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+CHANGED: `Select` component now exposes a `justValue` prop that can be bound to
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-select-just-value/?path=/story/ui-select--binidng-to-just-value

PR #281, opened 2024-04-16
Maps exports

jamesscottbrown:
This fixes a warning from `vite-plugin-svelte`:
> The following packages have a svelte field in their package.json
> but no exports condition for svelte

See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#missing-exports-condition

It does this by adding an exports condition (as for the other packages in the monorepo); MapLibre themes are now exported via index.js.

(see also https://github.com/Greater-London-Authority/ldn-viz-tools/commit/ae498d91358af59dd9faafcbe301a577e905ccb9)

### DIFF
--- a/.changeset/kgierif-dmiqwd-keiwie.md
+++ b/.changeset/kgierif-dmiqwd-keiwie.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/maps': major
+---
+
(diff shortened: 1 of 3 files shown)


ChrisKnightLDN:
LGTM

Issue #282, opened 2024-04-16
Add modalWidth to SidebarHeader or allow a SidebarHint to be slotted in


PaulioRandall:
**Is your feature request related to a problem? Please describe.**

`<SidebarHint>` allows for a `modalWidth` prop. The current `<SidebarHeader>` passes modal parameters to it but currently the `modalWidth` prop is not passed.

**Describe the solution you'd like**

Either:
1. Add `modalWidth` prop to `<SidebarHeader>` that gets passed to the internal `<SidebarHint>`.
2. Modify `<SidebarHeader>` so it accepts a `<SidebarHint>` as a slot circumventing the need to pass props along.
3. Or perhaps pass in and use `<svelte:component>`

PR #284, opened 2024-04-17
Fix/sidebar modal width

ChrisKnightLDN:
**What does this change?**
Sidebar elements that can display a hint now accept the sidebarHint component into a hint slot

**Why?**
Allows greater flexibility and reduces the responsibility surface of the parent component

**How?**
Slots over props

**Related issues**:
#282 

no

**How is it tested?**
Storybook

**How is it documented?**
Storybook

na

**Is it complete?**

### DIFF
--- a/.changeset/good-shrimps-change.md
+++ b/.changeset/good-shrimps-change.md
@@ -1,0 +1,5 @@
+---
+"@ldn-viz/ui": minor
+---
+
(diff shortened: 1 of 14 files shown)


ChrisKnightLDN:
My proposed solution is to set the tooltip `strategy: 'fixed'` which resolves the issue.

PR #298, opened 2024-05-07
fix storybook controls so that they update component in default stories

jamesscottbrown:
This fixes storybook controls so that they update the components in defult stories.

For the controls to work, the default `<Story>` tag must be empty - it's contents should instead be set in the `<Template>`.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-fix-storybook-controls/?path=/docs/ui-introduction--documentation

PR #299, opened 2024-05-07
Increase text size on tooltips

PaulioRandall:
**What does this change?**

Increases text size to `text-sm` for tooltip popup content.

**Why?**

Because current size is too small. Project clients requested it be increased for accessibility.

**How is it documented?**

Stories.

**Is it complete?**

### DIFF
--- a/.changeset/ghjghj_sdf_gfhio.md
+++ b/.changeset/ghjghj_sdf_gfhio.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+Increase Tooltip component text size for accessibility.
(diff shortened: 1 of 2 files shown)


.changeset/ghjghj_sdf_gfhio.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+Increase Tooltip component text size for accessibility.
```
jamesscottbrown, line 5:
```suggestion
CHANGED: Increase Tooltip component text size for accessibility.
```

Issue #295, opened 2024-05-03
Error state of Select does not reset correctly


jamesscottbrown:
On the `<Select>` component (and presumably any component using the same InputWrapper), setting the `error` prop to `true` makes the text of the label red, and applies a red border to the input itself.

If the prop is then changed to `false`, then the color of the label updates correctly, but the red border remains.

With `error={true}`:

![image](https://github.com/Greater-London-Authority/ldn-viz-tools/assets/338833/0760b4be-e7e4-422a-b524-c227ec06bc5f)

After changing to `error-{false}`:

![image](https://github.com/Greater-London-Authority/ldn-viz-tools/assets/338833/88dd2d60-403f-4ca4-b0c4-42c7ab7110e9)

jamesscottbrown:
This was resolved in https://github.com/Greater-London-Authority/ldn-viz-tools/pull/311

PR #301, opened 2024-05-07
Remove `hasError` prop from `Select`

jamesscottbrown:
This modifies `Select` so that it only has an `error` prop (rather than an `error` and `hasError` props)

This should address https://github.com/Greater-London-Authority/ldn-viz-tools/issues/295

@PaulioRandall, I don't know whether this touches on your planned changes to the `Select`

### DIFF
--- a/.changeset/dscsdf_dda_swwd.md
+++ b/.changeset/dscsdf_dda_swwd.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': major
+---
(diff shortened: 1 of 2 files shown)


PaulioRandall (approved):
LGTM. Regarding the use of `error` as indicator for `hasError`. Yes, I was planning on making the same change to the `Input` and `InputWrapper` component as well.

PR #306, opened 2024-05-07
Align text and icon in tooltip trigger

jamesscottbrown:
This changes the alignment of the icon in the trigger of the `<Tooltip>` component.

Before:

image

After:

image


ChrisKnightLDN:
@jamesscottbrown this is an improvement over what looks obviously wrong. I think we may wish to finesse further, tbd, but I'm happy to merge this.

PR #309, opened 2024-05-07
Node mailer util

PaulioRandall:
**What does this change?**

Adds the `postMail` function to the utils package that posts mail via node-mailer instance.

**Why?**

Because the API is easy to forget and error handling is non-standard. It's doubles up as documentation for sending requests to node-mailer.

**How is it tested?**

Used and tested in https://dev.ldn-gis.co.uk/cool-spaces/

**How is it documented?**

TypeScript and file comments.

**Is it complete?**


PaulioRandall:
Not needed, push to template instead.

PR #311, opened 2024-05-08
Update Input and InputWrapper

PaulioRandall:
**What does this change?**

Updates: 

1. `Input` to allow textarea type.
2. `Input` to pass `$$restProps` to the input element to account for differences in properties for different input types.
3. `Input` to export the bindable input element value.
4. `Input` to allow a format value function to be passed that formats the value when unfocusing the input.
5. `InputWrapper` to allow tooltip hints in the label.
6. `Input` and `InputWrapper` with aria and accessibility attributes.

**Why?**

So the input is usable and accessible with HTML forms and multiple input types.

**How is it tested?**

Storybook

**How is it documented?**

Stroybook

**Is it complete?**

### DIFF
--- a/.changeset/a_b_c.md
+++ b/.changeset/a_b_c.md
@@ -1,0 +1,10 @@
+---
+'@ldn-viz/ui': major
+---
+
+CHANGED: update `Input` to allow textarea type.
+CHANGED: update `Input` to pass `$$restProps` to the input element to account for differences in properties for different input types.
+CHANGED: update `Input` to export the bindable input element value.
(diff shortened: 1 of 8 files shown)


PaulioRandall:
@jamesscottbrown I just realized I need to update other components using the `<InputWrapper>`. Will let you know when I've done that.

PaulioRandall:
> @jamesscottbrown I just realized I need to update other components using the `<InputWrapper>`. Will let you know when I've done that.

@jamesscottbrown Good to review now.

jamesscottbrown:
This looks ok - I'll add some docs then merge.

PR #315, opened 2024-05-09
replace use of deprecated <Meta> tag in stories

jamesscottbrown:
This replaces the use of the deprecated `<Meta>` tag in some some stories with the now preferred approach of exporting a constant named  `meta`.

This updates the `.stories.svelte` files for UI components.


PaulioRandall (approved):
LGTM

PR #319, opened 2024-05-09
Add placeholder image component

PaulioRandall:
**What does this change?**

Adds the `<PlaceholderImage>` component.

**Why?**

A good placeholder image will fill a space where an image is expected but not available at page or widget load. This can be used to avoid an image, that was lazy loaded, changing the layout of a page or widget after presentation. This can be jarring for some users.

**How is it tested?**

Storybook

**How is it documented?**

Storybook

Yes

**Is it complete?**

### DIFF
--- a/.changeset/place-holder-image.md
+++ b/.changeset/place-holder-image.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+ADDED: `PlaceholderImage` component to show while waiting for the real image to load or where an image is expected but not available.


jamesscottbrown:
I've split out the classes to give a separate `strokeClass` prop.

Simply concatenating the default `stroke-core-blue-600` class with another user-specified class that sets the stroke color doesn't always give the intended result of overriding the defualt stroke color with the user-specificed one. Which one wins depends not on the order in which the classes are applied to the element, but rather on the order in which they are defined (the one that was defined last wins). This is a problem we enountered previously elsewhere in the component library.

PR #320, opened 2024-05-09
[docs] Select docs

jamesscottbrown:
This adds some documentation for the `<Select>` component


jamesscottbrown:
Current: https://dev.ldn-gis.co.uk/storybook-dev/?path=/docs/ui-select--documentation

Preview: https://dev.ldn-gis.co.uk/storybook-select-docs/?path=/docs/ui-select--documentation

PaulioRandall (approved):
Approved the docs thus far. I see a few bits missing but we can fill in later.

However, I think the `<Select>` component API and implementation needs organizing and slimming down as a separate task #328.

PR #325, opened 2024-05-10
[docs] add more docs for Header components

jamesscottbrown:
This adds some more docs for the header components.

It isn't perfect - for some reaosn Storybook doesn't seem to pick up `@component` comments for components with no props - but it is an improvement on the current docs.

### DIFF
--- a/packages/ui/src/lib/header/Header.svelte
+++ b/packages/ui/src/lib/header/Header.svelte
@@ -1,5 +1,22 @@
+<!--
+@component
+This is a test...
+-->
+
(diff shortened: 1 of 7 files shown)


packages/ui/src/lib/header/Header.svelte
```diff
@@ -1,5 +1,22 @@
+<!--
+@component
+This is a test...
+-->
```
PaulioRandall, line 4: Does it work? Does it get overwritten by docs in the script tag?

PR #327, opened 2024-05-10
[docs] add docs for sidebar

jamesscottbrown:
This adds some initial docs for the sidebar.

### DIFF
--- a/packages/ui/src/lib/sidebar/Sidebar.svelte
+++ b/packages/ui/src/lib/sidebar/Sidebar.svelte
@@ -33,6 +33,9 @@
 	 * position (usually the `AppShell` component).
 	 */
 	export let position: 'fixed' | 'absolute' = 'absolute';
+	/**
+	 * Colour scheme to use, either `light` or `dark`.
+	 */
 	export let theme: 'light' | 'dark' = 'dark';
 	export let placement: PlacementType = 'right';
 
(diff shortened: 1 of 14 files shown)


packages/ui/src/lib/sidebar/Sidebar.svelte
```diff
@@ -15,6 +22,9 @@
 	import { sidebarWidthStore } from './stores';
 	import type { PlacementType } from './types';
 
+	/**
+	 * Width of the sidebar.
+	 */
 	export let width: 'standard' | 'wide' = 'standard';
 	export let position: 'fixed' | 'absolute' = 'absolute';
```
PaulioRandall, line 29:
```suggestion
	/**
	 * Position of the sidebar which can be `fixed` to the screen so it's always shown
	 * in the same screen location regardless of scroll position or `absolute`, 
	 * positioned relative to the nearest parent  container with the `relative` CSS
	 * position (usually the `AppShell` component).
	 */
	export let position: 'fixed' | 'absolute' = 'absolute';
```

packages/ui/src/lib/sidebar/Sidebar.svelte
```diff
@@ -15,6 +22,9 @@
...
 	export let theme: 'light' | 'dark' = 'dark';
```
PaulioRandall, line 30:
```suggestion
	/**
	 * Colour scheme to use, either `light` or `dark`.
	 */
	export let theme: 'light' | 'dark' = 'dark';
```

PaulioRandall (approved):
LGTM. Proposed docs for two missing props that you may want to include.

PR #340, opened 2024-05-16
add helper function and tweak example plots to match those in PDF report

jamesscottbrown:
Added `preprocessOptions()` function for creating multiple marks with different values of options in Observable Plot.

This make sit easier to create a set of labels with different offsets from the corrresponding data points.
The usefulness of this is shown by the example storeis,w hich now have label placement much closer to those in the PDF report that they try to replicate.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-observable-plot-annotations/?path=/docs/charts-observableplotfragments--documentation

PR #346, opened 2024-05-17
WIP: start refactor of modals and tooltips

jamesscottbrown:
The idea is that each "help provider" (tooltip, popover, modal) can render both its "trigger" and it's content.

There is a standard `Trigger` component: it receives a context containing functions to call when it is interacted with.

This is a very incomplete WIP that I've added as a draft PR so I don't forget the idea.


jamesscottbrown:
This has been superceeded by https://github.com/Greater-London-Authority/ldn-viz-tools/pull/837

PR #350, opened 2024-05-22
Remove MapApp component

PaulioRandall:
**What does this change?**

Removes the `<MapApp>` component that was used as an application shell for maps.

**Why?**

It's been replaced by `<AppShell>`.

**How?**

Removed from _packages/maps/src/lib/index.js_.


PaulioRandall:
Out of date.

PR #355, opened 2024-06-07
Update Map with map cursor

PaulioRandall:
**What does this change?**

1. Updates the `<Map>` component with TypeScript.
2. Renames `map` context in `<Map>` to `mapStore`.
3. Adds functionality for effective of handling map cursor events such as hovers & clicks including adding `mapCursorStore` context.
4. Adds props `mapStore` and `mapCursorStore` to `<Map>` to enable reactivity based map loading.

**Why?**

The way MapLibre uses JavaScript's event system makes it difficult and verbose to manage propagation of events within project specific code. Specifically, ensuring events for top most features take precedence in controlling the state of the map including current popup, highlights, mouse style, or any other map or app state. For example:
- A `mouseenter` event setting the cursor maybe overridden in the same tick by a `mouseleave` event for a different feature, possible in an unrelated layer. The top most feature should determine the final cursor style.
- Clicking multiple features in the same location should only show a popup (`marker`) for the top most feature but we also want to keep popup specific code with its layer definition and isolated from other popups.

**How is it tested?**

Storybook and in live applications.

**How is it documented?**

Storybook.

No.

**Is it complete?**


PaulioRandall:
@jamesscottbrown I've just noticed a small issue I introduced in the last commit. Hold off on this one.

PaulioRandall:
> @jamesscottbrown I've just noticed a small issue I introduced in the last commit. Hold off on this one.

@jamesscottbrown Good to go.

packages/maps/package.json
```diff
@@ -35,6 +35,7 @@
 		"@sveltejs/adapter-static": "^3.0.1",
 		"@sveltejs/package": "^2.3.0",
 		"@sveltejs/vite-plugin-svelte": "^3.0.2",
+		"@turf/turf": "6.5.0",
```
jamesscottbrown, line 38: Why is this import added?

packages/maps/src/lib/map/Map.svelte
```diff
@@ -1,25 +1,46 @@
...
+	} from '@ldn-viz/maps';
+	import MapCursor from './mapCursor/MapCursor';
+	import type { MapCursorStore } from './mapCursor/types';
+
+	export type MapLibre = maplibre_gl.Map;
+	export type MapStore = Writable<null | maplibre_gl.Map>;
+	export type WhenMapLoads = (map: maplibre_gl.Map) => void;
+
+	/**
+	 * Apllied as the MapLibre transformRequest option to append the OS key
```
jamesscottbrown, line 20:
```suggestion
	 * Applied as the MapLibre `transformRequest` option to append the OS key
```

packages/maps/src/lib/map/mapCursor/HandlerCollection.js
```diff
@@ -0,0 +1,120 @@
...
+				events.splice(i, 1);
+				return true;
+			}
+		}
+
+		return false;
+	};
+
+	const updateLayerIds = () => {
+		const nonUniqueLayerIds = Object.values(self.entries) //
```
jamesscottbrown, line 113:
```suggestion
		const nonUniqueLayerIds = Object.values(self.entries)
```

PR #363, opened 2024-06-17
group map control component stories together

jamesscottbrown:
This groups together the stories for the different map control components into one directory in the sidebar of the Storybook site:

image


ChrisKnightLDN:
Can this folder be called MapControl for consistency

PR #366, opened 2024-06-21
Design tokens

ChrisKnightLDN:
Major change updates theme and EVERY COMPONENT to use design tokens and css vars for theme switching etc.


jamesscottbrown:
Preview of how this looks in Storybook: https://dev.ldn-gis.co.uk/storybook-design-tokens/?path=/docs/introduction--documentation

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+
+const conditionsCssDarkTheme = [{ category: 'theme', type: 'dark' }];
+
+const conditionsCssGlobal = [
+  // token.attributes.category === 'global' || token.attributes.category === 'spacing'
+  { category: 'global', subitem: { not: ['seed'] } },
+  { category: 'spacing' }
+];
+
+const conditionsTw = [
```
jamesscottbrown, line 112: I'd rename this to `conditionsTailwind`: it is more explicit, and not too long.

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+
+const matchesCondition = (token, condition) => {
+  return Object.entries(condition).every(([key, value]) => {
+    if (typeof value === 'object' && value.not) {
+      return !value.not.includes(token.attributes[key]);
+    }
+    return token.attributes[key] === value;
+  });
+};
+
```
jamesscottbrown, line 136: Renaming `[key, value]` to `[field, filter]` and extracting a `val` constant is possibly slightly clearer:

```suggestion
const matchesCondition = (token, condition) => {
  return Object.entries(condition).every(([field, filter]) => {
    const val = token.attributes[field];
    
    if (typeof filter === 'object' && filter.not) {
      return !filter.not.includes(val);
    }
    return filter === val;
  });
};
```

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+const tokenMatchesAnyCondition = (token, conditions) => {
+  return conditions.some((condition) => matchesCondition(token, condition));
+};
+
+const createMatcher = (conditions) => (token) => tokenMatchesAnyCondition(token, conditions);
+
+// REGISTER THE CUSTOM FILTERS USING OUR MATCHING CONDITIONS
+
```
jamesscottbrown, line 144: I'd move the definitions of the conditions objects to just before they are used here (I'd define `conditionsCssGlobal`, then use it register a filter, then define `conditionsCssLightTheme`, then use that to register a filter ....)

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+});
+
+/**
+ * Custom format that handle reference in css variables
+ */
+const transformString = (str, replace = 'theme', regex = /.*-color/) => {
+  // Use a regular expression to replace everything before regex with replace
+  return str.replace(regex, replace);
+};
+
```
jamesscottbrown, line 177: This is misleading: the function replaces the regex match, **not** everything before the regex match.
(of course, if the regex begins with `.*`, then it that part of the regex will match any amount of anything before the rest of the match, but whatever it matches is part of the regex match).

All this function does is supply default values to `str.replace`: I'd remove it and use `str.replace()` directly. This has the advantage of being more explicit about what substitutions are being done.

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+                return `  --${themedName}: var(--${referenceName}, ${token.value});`;
+              }
+
+              return `  --${themedName}: ${token.value};`;
+            })
+            .join('\n')}
+        }`;
+  }
+});
+
```
jamesscottbrown, line 206: There is too much nesting in this template string - I'd extract out the anonymous function:

```suggestion

const formatCSSVariable = (token) => {
    let originalName = token.name;

    let themedName =
      token.path[0] === 'global'
        ? originalName
        : transformString(originalName, 'theme-', /.*-color-/);

    if (dictionary.usesReference(token.original.value)) {
      const reference = dictionary.getReferences(token.original.value);
      const referenceName =
        reference[0].path[0] === 'global'
          ? reference[0].name
          : transformString(reference[0].name);
      return `  --${themedName}: var(--${referenceName}, ${token.value});`;
    }

    return `  --${themedName}: ${token.value};`;
}

StyleDictionary.registerFormat({
  name: 'css/variables',
  formatter({ dictionary }) {
    return `${this.options.selector} {
          ${dictionary.allTokens.map(formatCSSVariable).join('\n')}
        }`;
  }
});

```

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+              : transformString(originalName, 'theme-', /.*-color-/);
+
+          let transformedName = transformString(originalName, 'color-', /.*-color-/);
+
+          return `  "${transformedName}": "var(--${themedName}, ${token.value})"`;
+        })
+        .join(',\n')}\n}`
+    );
+  }
+});
```
jamesscottbrown, line 232: Again, I'd extract the anonymous function out of the template string:


```suggestion

const formatTailwindColor = (token) => {
    let originalName = token.name;

    let themedName =
      token.path[0] === 'global'
        ? transformString(originalName, 'global-color')
        : transformString(originalName, 'theme-', /.*-color-/);

    let transformedName = transformString(originalName, 'color-', /.*-color-/);

    return `  "${transformedName}": "var(--${themedName}, ${token.value})"`;
}

StyleDictionary.registerFormat({
  name: 'tw/css-variables',
  formatter({ dictionary }) {
    return (
      'module.exports = ' +
      `{\n${dictionary.allTokens.map(formatTailwindColor).join(',\n')}\n}`
    );
  }
});
```

Also, you can use a template string instead of the string concatenation:

```js
'module.exports = ' +
`{\n${dictionary.allTokens.map(formatTailwindColor).join(',\n')}\n}`
```

```js
`module.exports = {
  ${dictionary.allTokens.map(formatTailwindColor).join(',\n')}
}
`
```

jamesscottbrown (commented):
I've made some comments to try and make it easier to understand what `build.js` does - it applies fairly complicated transformations.

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+    category: 'global',
+    type: { not: ['spacing'] },
+    item: { not: ['mode'] },
+    subitem: { not: ['seed'] }
+  }
+];
+
+const matchesCondition = (token, condition) => {
+  return Object.entries(condition).every(([key, value]) => {
+    if (typeof value === 'object' && value.not) {
```
jamesscottbrown, line 130: I'm not sure what should happen if `value` is an object but does not have a `not` attribute - throw an error?

packages/themes/build.js
```diff
@@ -0,0 +1,243 @@
...
+};
+
+StyleDictionary.registerParser({
+  pattern: /\.json$/,
+  parse: ({ contents }) => {
+    try {
+      const jsonObj = JSON.parse(contents);
+
+      // Replace the references to modes
+      const lightParsed = replaceStringInMode(
```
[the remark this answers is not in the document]
ChrisKnightLDN: yeah - it's a work around of a bug

packages/themes/tokens/design-tokens.tokens.json
```diff
@@ -0,0 +1,4439 @@
...
+            "yellow": {
+              "description": "reference only - do not use",
+              "type": "color",
+              "value": "{global.color.static.gla.yellow}"
+            },
+            "turquoise": {
+              "type": "color",
+              "value": "#63c5b5ff",
+              "blendMode": "normal"
+            }
```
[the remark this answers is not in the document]
ChrisKnightLDN: Have updated

PaulioRandall (commented):
I think this PR is too big. Difficult to review but most components look fine in Storybook. 

The contrast on the map controls is off though. You can probably resolve this by forcing dark mode in `<Map>` or I can force dark mode in each map app. We can resolve map color schemes independently:

image

packages/charts/src/lib/chartContainer/SubTitle.svelte
```diff
@@ -1 +1 @@
-<div class="text-core-grey-500 dark:text-core-grey-200"><slot /></div>
+<div class="text-color-text-secondary"><slot /></div>
```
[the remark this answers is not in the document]
ChrisKnightLDN: this is inherent in a token structure. The token is "color-text-secondary" - the initial application is the tailwind directive "text" hence "text-color-text-secondary" I'd argur that this is specificity, not redundancy

ChrisKnightLDN:
> The contrast on the map controls is off though. You can probably resolve this by forcing dark mode in `<Map>` or I can force dark mode in each map app. We can resolve map color schemes independently:
> 
> image

yes - given the ongoing work on map components I held off addressing this - a global dark class would revert the buttons to their current dark grey apperence.

packages/themes/build.js
ChrisKnightLDN: ok

jamesscottbrown:
These changes are now in the `next` branch.

PR #377, opened 2024-06-27
simplify merge values control

jamesscottbrown:
This simplifies the `MergeValuesControl` component. There's potential for further design tweaing, but this is an imprvement.


ChrisKnightLDN:
Yes - needs design love - could you modify the list items so that they don't change size on hover at least (as it causes page-jank on hover) We can then review more fully in future design iteration

jamesscottbrown:
> could you modify the list items so that they don't change size on hover at least (as it causes page-jank on hover)

Done.

Issue #378, opened 2024-06-27
Build warning annotations


ChrisKnightLDN:
Minor annoyance is warning annotations on build etc. feel like these should be addressed and not ignored:
![image](https://github.com/Greater-London-Authority/ldn-viz-tools/assets/111759008/6f2bab7d-aefb-4fee-a427-25bb09d41091)

PR #379, opened 2024-06-28
update versions of actions in workflows

jamesscottbrown:
This updates the version of actions used in our GitHub Action workflows, which addresses one of the warnings flagged in https://github.com/Greater-London-Authority/ldn-viz-tools/issues/378


ChrisKnightLDN:
lgtm

PR #383, opened 2024-06-28
Unfix sidebar header

PaulioRandall:
**What does this change?**

Unfix the sidebar header so it scrolls with the section content

**Why?**

Because on small screens it leaves next to no real estate for the content. Users find it hard to see content.

**Is it complete?**


ChrisKnightLDN (commented):
I think this is ok as a quick fix. We may want to look at the behaviour over a range of breakpoints - but this will address the issue.

PR #409, opened 2024-07-22
Borough map control

jamesscottbrown:
**What does this change?**

This adds a map component that zooms to a borough selected from a select box.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-borough-control/?path=/docs/maps-mapcontrols-mapcontrolborough--documentation

PR #411, opened 2024-07-23
Schema driven forms

PaulioRandall:
**What does this change?**

Adds components for constructing a form via a schema rather than straight props:

- `SchemaForm` component for creating a `Form` from a schema.
- `SchemaFormField` (internal) used by `SchemaForm` that renders both library and custom input types.

**Why?**

So that forms can be driven via objects which provides more modularity and conciseness at the expense of customization.

No.

**How is it tested?**

Storybook.

**How is it documented?**

Storybook.

No.

**Is it complete?**


jamesscottbrown:
Currently, in the "Simple Form" story, filling in the form then pressing "Clear" removes the `color` key from the value sobject.

PaulioRandall:
> Currently, in the "Simple Form" story, filling in the form then pressing "Clear" removes the `color` key from the value sobject.

Hmm, seems to be a quirk of the `<Select>` component. I think the _Select_ component has issues being reactively updated via `justValue`. I'll create an issue.

It shouldn't affect how the forms work since a missing key is the same as being `undefined` or an empty string.

PR #412, opened 2024-07-23
allow SidebarSection component to accept a SidebarSectionTitle compon…

jamesscottbrown:
This changes the `<SidebarSection>` component so that it can accept a `<SidebarSectionTitle>` component inside the `title` slot. This allows the use of a subtitle and help/hint text.

### DIFF
--- a/.changeset/sfsdf-fsdfsdf-fsdfsdf.md
+++ b/.changeset/sfsdf-fsdfsdf-fsdfsdf.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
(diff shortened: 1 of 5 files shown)


ChrisKnightLDN:
Have adjusted typography to better reflect figma. @jamesscottbrown do you think the title prop AND the slot are required - should it be one way or another to avoid confusion?

jamesscottbrown:
We have a similar pattern for customization elsehwere (e.g., the `Popover`, where the trigger can be customized slightly by setting the `hintLabel` prop, or customized more by populating the `hint` slot).

Issue #407, opened 2024-07-19
Input component not styled correctly when type=range


jamesscottbrown:
The top image is an HTML `input` element; the lower image is our `Input` component. Both have `type='range'`, so that they are rendered as sliders.

![image](https://github.com/user-attachments/assets/e5c0ec6e-1732-4269-be53-3353962d6468)

We plan to add a separate `Slider` contol with more functionality, but it would still be good for the `Input` component to handle the case of a basic slider.

PR #425, opened 2024-07-24
fix styling on Input element when type=range

jamesscottbrown:
**What does this change?**

Fix styling of `Input` element when `type='range'`

See https://github.com/Greater-London-Authority/ldn-viz-tools/issues/407

**Is it complete?**

### DIFF
--- a/.changeset/fix-range-input.md
+++ b/.changeset/fix-range-input.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
See preview: https://dev.ldn-gis.co.uk/storybook-fix-input-slider/?path=/story/ui-input--alternative-types

ChrisKnightLDN:
@jamesscottbrown I updated the range to accept a class that styles the blue to use our blue - not the default. If you are happy this can be merged

Issue #133, opened 2023-10-06
Button loading spinner


ChrisKnightLDN:
There is a design for a button with a loading state inc. spinner

jamesscottbrown:
We now have:
* the `AsyncButton` component: https://greater-london-authority.github.io/ldn-viz-tools/?path=/docs/ui-asyncbutton--documentation
* a `Spinner` component, with a story showing it being used in a button: http://localhost:6006/?path=/story/ui-spinner--in-a-button

Is there anything still to do here?

ChrisKnightLDN:
@jamesscottbrown I think we just need to align these - the async button does not provide ui text for its state ie. "loading" and the button with spinner does not provide much in the way of functionality. Mostly a case of removing ambiguity

PR #448, opened 2024-08-01
Add AsyncButton

PaulioRandall:
**What does this change?**

Adds the `<AsyncButton>` component and changes `<FormSubmitButton>` to use it instead of the `/forms` internal version.

**Why?**

1. Disables the button preventing multiple clicks for slow external requests.
2. Adds a spinner to tell the user something is happening so please stand by.
3. Provides a `working` prop so that other parts of the application know when the click operation is executing and when it stops.

**How?**

Wraps `<Button>` but requires the handler to be passed to `onClick` prop.

**How is it tested?**

Storybook

**How is it documented?**

Storybook

No

**Is it complete?**


ChrisKnightLDN:
@PaulioRandall We should be able to adress the thickness of the spinner: https://greater-london-authority.github.io/ldn-viz-tools/?path=/docs/ui-spinner--documentation#thickness. Apart from that this looks ok to me styling wise for now - as discussed

ChrisKnightLDN:
related: #133

PR #449, opened 2024-08-01
use Popover component from ui package rather than PopoverMenu component in table package 

jamesscottbrown:


ChrisKnightLDN:
LGTM - caveated with our global review of hints -> popover, modal, tooltip

Issue #333, opened 2024-05-13
Geocoder should display name of selected location, not filtering string


jamesscottbrown:
Aften selecting a location form the drop-down list provided by the geocoder, the select field should contain the name of the location (rather than just the string typed by the user).

For example, if I type "Unio" to filter the list, then click to accept a suggestion form the list, then"Unio" should be replaced by the full address ("Union Street, Southwark, London, SE1 1SG, United Kingdom"). This is the behaviour of `@mapbox/mapbox-gl-geocoder`.

PR #484, opened 2024-08-02
Set selected location in Geocoder as query text 

PaulioRandall:
**What does this change?**

Updates `<Geocoder>` so that the input's text is the address or name of the selected location when a location is selected.

**Why?**

So users know what they selected.

**Related issues**:

Addresses: #333 

**How is it tested?**

Storybook.

**How is it documented?**

Storybook.

**Is it complete?**

### DIFF
--- a/.changeset/geocoder-selected-name.md
+++ b/.changeset/geocoder-selected-name.md
@@ -1,0 +1,6 @@
+---
+'@ldn-viz/ui': minor
+---
+
+CHANGED: `Geocoder` selected location now appears as query text on selection.
+ADDED: adds reactively updating `selected` property to `Geocoder` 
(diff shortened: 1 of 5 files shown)


PaulioRandall:
**Preview:** https://dev.ldn-gis.co.uk/storybook-geocoder-selected-name/?path=/docs/ui-geocoder--documentation

PaulioRandall:
> I think it should set the text to be the _name_ rather than _address_ of the selected feature.
> 
> For example, if I select the "Sir John McDougall Gardens" in the example, I'd want the input box to display "Sir John McDougall Gardens" (rather than "Westferry Rd, Tower Hamlets, London"), since the name is what I searched for and selected.

@jamesscottbrown I've amended the suggestion list so the name is shown on select. It also highlights the currently selected item when reopening the suggestion list.

PaulioRandall:
**Preview:** https://dev.ldn-gis.co.uk/storybook-geocoder-selected-name/?path=/docs/maps-mapcontrols-mapcontrollocationsearch--documentation

Issue #370, opened 2024-06-25
Input component does not grey out text when disabled


PaulioRandall:
**Describe the bug**

`<Input disabled />` does not have greyed out text when disabled so it looks editable. However, the input label is greyed out.

**To Reproduce**

Write the following Svelte code:

```js
import { Input } from '@ldn-viz/ui'

<Input name="meh" value="some text" disabled />
```

**Expected behavior**

The input text should be greyed out the same as the label text.

jamesscottbrown:
The input box itself is greyed out: if the text inside the box was also greyed out it would be difficult to provide enough contrast between the text and background.

How the component currently looks in the disabled state:
![Screenshot 2024-06-26 at 12 10 18](https://github.com/Greater-London-Authority/ldn-viz-tools/assets/338833/12ff4907-4e8e-490a-ad9f-fddab6e8057d)

PaulioRandall:
> The input box itself is greyed out: if the text inside the box was also greyed out it would be difficult to provide enough contrast between the text and background.
> 
[...]

PR #492, opened 2024-08-06
Fix input text color

ChrisKnightLDN:
**What does this change?**
fixes color of input text when disabled

**Why?**
it was incorrect before

**How?**
adjust prop and add color classes

**Related issues**:
#370 

no

**How is it tested?**
sb

**How is it documented?**
na

yes

**Is it complete?**


packages/ui/src/lib/input/Input.svelte
jamesscottbrown, line 1: Does line 132 need to be updated in the same way?

> disabled: !!$$restProps.disabled

Issue #498, opened 2024-08-08
Themes package not exposing built css files


ChrisKnightLDN:
**Describe the bug**
On installation into an app I need to be able to reference the css files generated on build in themes. These are not available.

<img width="162" alt="Screenshot 2024-08-08 at 10 52 39" src="https://github.com/user-attachments/assets/28fc067a-e816-4d40-be15-13e770b2600e">

I would expect the 'styles' directory to be present and to be able to reach the files inside

PR #499, opened 2024-08-08
add .npmignore file for themes package

jamesscottbrown:
**What does this change?**

Adds a `.npmignore` file - without this, npm exlcudes the files listed in `.gitignore`, so the generated CSS files aren't included in the published package and can't be accessed from other projects.


ChrisKnightLDN:
resolves: #498

PR #552, opened 2024-09-02
Map dark mode

PaulioRandall:
**What does this change?**

Adds dark mode to `Map` component including new properties to set light and dark map styles.

**How?**

A `style` value must no longer be provided to the `options` prop.

Adds props:
- `lightStyle`: light map style to use or null to disable light mode.
- `darkStyle`: dark map style to use or null to disable dark mode.

**How is it tested?**

Storybook.

**How is it documented?**

Storybook.

Yes.

**Is it complete?**

### DIFF
--- a/.changeset/map-dark-mode.md
+++ b/.changeset/map-dark-mode.md
@@ -1,0 +1,7 @@
+---
+'@ldn-viz/maps': major
+---
+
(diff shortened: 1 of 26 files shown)


PaulioRandall:
Wrong branch

PR #555, opened 2024-09-02
Rewrite map control group & fix its types

PaulioRandall:
**What does this change?**

Rewrites `MapControlGroup` interface to allow finer grained control over positioning.

**Why?**

As the number of artifacts on the page grows the positioning of additional map controls was becoming difficult, e.g. map theme switcher.

**How is it tested?**

Storybook.

**How is it documented?**

Storybook.

No.

**Is it complete?**

### DIFF
--- a/packages/maps/src/lib/mapControlGroup/MapControlGroup.stories.svelte
+++ b/packages/maps/src/lib/mapControlGroup/MapControlGroup.stories.svelte
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
 	import { Meta, Story, Template } from '@storybook/addon-svelte-csf';
 
 	import Map, { appendOSKeyToUrl } from '../map/Map.svelte';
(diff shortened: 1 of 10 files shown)


packages/maps/src/lib/mapControlGroup/MapControlGroup.stories.svelte
```diff
@@ -22,78 +40,97 @@
...
+<Story name="Default" source />
+
+<Story name="Positions">
+	<div class="w-[100dvw] h-[100dvh] bg-color-bg-primary relative">
+		{#each yPositions as yPos (yPos)}
+			{#each xPositions as xPos (xPos)}
+				<MapControlGroup y={yPos} x={xPos}>
+					<p class="bg-color-container-level-1 text-color-text-primary p-2 capitalize">
+						{yPos}
+						{xPos}
```
jamesscottbrown, line 68: Would be clearer to display as something like `x={xPos} y={yPos}` or `y={yPos} x={xPos}`

PR #560, opened 2024-09-02
Fix map control pan types

PaulioRandall:
**What does this change?**

Fix TypeScript for `MapControlPan`.

**Is it complete?**


jamesscottbrown:
I think we'll leave the positions specified as they currently are. We'll check the types of the pan controls when checking the types of everything in the library in the new year.

PR #570, opened 2024-09-18
Add function to generate simple diverging colors

halimahLDN:
`getThresholdBreaksColorsLabels()` has a bug that makes it difficult to get the right count of colors with `extent` and `breakCount`, and is rigid in the structure of the arguments which limits its use for simpler color ramps.

I wrote this function for simple use cases where users want to pass in two arrays of colors to get a diverging color ramp at a specific count. It calls `getColorRamp` twice and concatenates the resulting arrays, inserting a neutral middle color if the count is an odd number.

**How is it tested?**
Pass in a count, arrays of colors and boolean to the function. It should return an array with a length equal to count.

You can see it in practice [here](https://github.com/Greater-London-Authority/app-friday-effect-svelte/tree/new-refactor-data-loading). Relevant functions are saved in `src/lib/utils/utils.ts`.

You can change the colors in `darkPinkOrange` and this will update the bar charts:
```ts
export const darkPinkOrange = () => {
	const lowColors = [
		ldnColors.core.darkPink[800],
		ldnColors.core.darkPink[500],
		ldnColors.core.darkPink[200]
	];
	const highColors = [
		ldnColors.core.orange[200],
		ldnColors.core.orange[500],
		ldnColors.core.orange[800]
	];

	return getSimpleDivergingColors(8, lowColors, highColors, true);
};
```
image

### DIFF
--- a/.changeset/get-simple-diverging.md
+++ b/.changeset/get-simple-diverging.md
@@ -2,4 +2,4 @@
 '@ldn-viz/utils': minor
 ---
 
-ADDED: `getSimpleDiverging()` simple function to generate diverging color range
+ADDED: `getSimpleDiverging()` simple function to generate diverging color range from two color arrays
--- a/.changeset/get-simple-diverging.md
+++ b/.changeset/get-simple-diverging.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/utils': minor
+---
+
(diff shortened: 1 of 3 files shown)


ChrisKnightLDN:
Closing as there is an underlying issue that this attempts to circumvent. It is better for us to fix the issue. see: #711

PR #575, opened 2024-10-02
Add map drawing component

jamesscottbrown:
This is a draft PR that shows how TerraDraw can be integrated to allow drawing features on top of a map.

However, more design work is needed before it is merged into the component library.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-map-drawing/?path=/docs/maps-mapdraw--documentation

PR #591, opened 2024-10-11
don't fix table height unnecessarily

jamesscottbrown:
This updates the `Table` to no longer set a fixed height unless the `virtualise` prop is `true`.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-table-height/?path=/docs/tables-table--documentation

PR #592, opened 2024-10-11
default to showColSummaries = false

jamesscottbrown:
`Table` now defaults to `showColSummaries` being `false`.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-default-no-col-summary/

PR #593, opened 2024-10-11
Table issues

ChrisKnightLDN:
**What does this change?**
updates the table container component to align with the API of the chart container
attempts to resolve various type issues


jamesscottbrown:
I want to separate out the chart wrapper API changes from the changes to types: https://github.com/Greater-London-Authority/ldn-viz-tools/pull/594 includes only the former

PR #594, opened 2024-10-14
TABLES: updates table container to align with the API of the chart container

jamesscottbrown:
Updates the table container component to align with the API of the chart container


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-table-chart-api/?path=/docs/tables-table--documentation

PR #595, opened 2024-10-14
Tables: allow rule above table header to be disabled

jamesscottbrown:
This allows the horizontal rule above the header of a table to be disabled, to match the design in the Figma mockups for the HSDS hub.


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-table-optional-top-rule/?path=/docs/tables-table-hsds--documentation

PR #596, opened 2024-10-14
Tables: allow a gap between column groups in tables

jamesscottbrown:


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-tables-col-group-gap/?path=/docs/tables-table-hsds--documentation

PR #602, opened 2024-10-15
Table: remove class that was accidentally applied to header

jamesscottbrown:


jamesscottbrown:
Preview will appear at: https://dev.ldn-gis.co.uk/storybook-fix-table-header-rule/?path=/docs/tables-table-visual-encodings--documentation

PR #625, opened 2024-10-21
Add Callout component

jamesscottbrown:
**What does this change?**

This adds the `Callout` component from the hsds-hub project

**Why?**

There is a need for a similar component in the LBSM project, and it is likely to also be needed in other future projects.

**Is it complete?**

### DIFF
--- a/.changeset/callout.md
+++ b/.changeset/callout.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+ADDED: add `Callout` component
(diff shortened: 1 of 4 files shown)


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-callout

jamesscottbrown:
I'm not sure about the choice of `note`/`caution`/`warning` as the values of the `status` prop, since `caution` and `warning` seem like synonyms: perhaps `note`/`warning`/`error` would be clearer?

Issue #663, opened 2024-11-25
Make hintLabel an optional prop and add option to change flex direction


halimahLDN:
`RadioButtonGroup` takes a prop `hintLabel` which should be optional but is [currently expected](https://github.com/Greater-London-Authority/ldn-viz-tools/blob/43ff53a12eb17c90acd7176d5132662c36c5464d/packages/ui/src/lib/radioButton/RadioButtonGroup.svelte#L38C3-L38C21) in all use cases:

```ts
// this:
hintLabel: string;

// should be:
hintLabel?: string | undefined;
```

And for use of radio buttons in chart controls, it would be useful to arrange them in a row, rather than a column. Would it be possible to add an optional prop for this?:

```ts
// inside RadioButtonGroup.svelte.d.ts

flexDirection?: 'flex-col' | 'flex-row' | undefined;

// inside RadioButtonGroup.svelte
export let flexDirection = 'flex-col';
```

PR #664, opened 2024-11-25
Radio group props

halimahLDN:
**What does this change?**
- RadioGroup options no longer expect `hintLabel` prop (which was the intended functionality)
- Added new `RadioButtonGroup` story to display radio buttons in a row, with new optional props: `flexDirection`, `marginX`, `marginY`
     - note, `marginX` and `marginY` are currently strings but could be numbers (I tried this, i.e. `space-x-{marginX}`, but it didn't work)
image

**Related issues**: #663 

**Is it complete?**

### DIFF
--- a/packages/ui/src/lib/radioButton/RadioButtonGroup.svelte
+++ b/packages/ui/src/lib/radioButton/RadioButtonGroup.svelte
@@ -39,19 +39,19 @@
 	}[] = [];
 
 	/**
-	 * Default `flex-direction` is `flex-col` but allows you to create horizontal radios
-	 * if you pass in an empty string (default `flex` behaviour is to display in row).
+	 * Direction of the flexbox. Default direction is column but passing in `flex-row` or an
+	 * empty string will allow you to display radios in a row.
 	 */
 	export let flexDirection = 'flex-col';
 
 	/**
-	 * Optional prop to change margin on the x-axis between elements.
+	 * Optional prop to change margin between elements on the x-axis.
 	 * For example `space-x-4`.
 	 */
 	export let marginX = '';
 
 	/**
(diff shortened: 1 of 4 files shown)


ChrisKnightLDN:
I'm not totally convinced this is required - at least not with this much granularity in the api (margins etc). Perhaps a layout: vertical/ horizontal? (hintlabel should be optional - that is a bug)

halimahLDN:
I've refactored this so the `hintLabel` change is in a separate PR (#666). And now `RadioButtonGroup` takes one prop `orientation` which changes the alignment of the radio buttons:
image

halimahLDN:
@ChrisKnightLDN I don't believe this impacts the behaviour of the tooltip. See below screenshot:
image

packages/ui/src/lib/radioButton/RadioButtonGroup.svelte
```diff
@@ -48,16 +57,24 @@
...
-			{name}
-		/>
-	{/each}
+	<div
+		class={orientation === 'vertical'
+			? 'flex flex-col space-y-0.25'
+			: orientation === 'horizontal'
+				? 'flex space-x-3'
+				: ''}
+	>
```
jamesscottbrown, line 66: Does this need to handle the case where orientation is neither `'horizontal'` or '`vertical'`, and is applying no classes a useful thing to do in this case?

 I think we could simplify to:

```suggestion
		class={orientation === 'vertical'
			? 'flex flex-col space-y-0.25'
			: 'flex space-x-3' }
	>
```

halimahLDN: That makes a lot of sense - thanks, James!

PR #687, opened 2024-12-04
Change sidebarAlways open object to use Booleans

jamesscottbrown:
It would be better to switch to *requiring* the use of Booleans rather than strings, but that is a breaking change that can be deferred for now.


jamesscottbrown:
I'll create a new PR that makes the breaking change of using Booleans rather than strings.

PR #724, opened 2025-01-02
shift map control groups to avoid overlap with attribution (fix #669)

### LINKED ISSUE #669: Maps: pan controls in BottomRight position overlap attribution statement

jamesscottbrown:
Before: https://greater-london-authority.github.io/ldn-viz-tools/?path=/story/maps-mapcontrols-mapcontrolgroup--positioning-labels

After: https://dev.ldn-gis.co.uk/storybook-map-control-group-positioning/?path=/story/maps-mapcontrols-mapcontrolgroup--positioning-labels

### DIFF
--- a/.changeset/map-control-positions.md
+++ b/.changeset/map-control-positions.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/maps': patch
+---
+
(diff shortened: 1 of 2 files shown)


ChrisKnightLDN:
@jamesscottbrown this looks a bit unbalanced - would it be better to mirror the topRightOffset for the bottom right? or provide a class override so that this can be dynamic and respect tailwind breakpoints where needed?

Issue #469, opened 2024-08-01
Table: Sorting icon should indicate sort direction


ChrisKnightLDN:
![Screenshot 2024-08-01 at 16 52 46](https://github.com/user-attachments/assets/a8616d67-c98f-46a0-ba77-d8b398982c6f)

jamesscottbrown:
Yes, it should use unambiguous sort-direction icons like those in Font Awesome (`sortAlphaUp`, `sortAlphaDown`, `sortAmountUp`, `sortAmountDown`)/Octicons (`sortAsc`, `sortDesc`)/ Boostrap (`sortAlphaUp`, `sortAlphaDown`, `sortDown`, `sortUp`)/Flat Color (`AlphabeticalSortingAz`, `AlphabeticalSortingAz`, `NumericalSorting12`, `NumericalSorting21`)/IcoMoon (`SortNumbericDesc`, `SortNumbericAsc`, `SortAlphaAsc`, `SortAlphaDsc`) /Line Awesome (`SortAlphaDownSolid`, `SortAlphaUpSolid`, `SortAmountDownSolid`, `SortAmountUpSolid`) - not just a simple up-down arrow.

Heroicons only has `arrow-up`/`arrow-down` and `chevron-up`/`chevron-down` icons, so we'll need to use something from outside this set (either borrowed from a different icon library or drawn custom).

ChrisKnightLDN:
Heroicons bars-arrow-up, bars-arrow-down

PR #725, opened 2025-01-02
change table sort icons to indicate applied sort direction

jamesscottbrown:
This fixes https://github.com/Greater-London-Authority/ldn-viz-tools/issues/469

### DIFF
--- a/.changeset/table-sort-icons.md
+++ b/.changeset/table-sort-icons.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/tables': patch
+---
+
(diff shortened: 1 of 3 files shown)


jamesscottbrown:
The bars and arrows should be a redundant encoding, with both showing the sort direction.

Octicons gets this right with `sort-asc`/`sort-desc`:

image

image





But Heroicons `bars-arrow-up`/`bars-arrow-up` change only the arrow direction, not the order of the bars:

image

image


Rather than conditinally rendering one of the pair of Heroicons icons, I'm using the same one and conditionally applying a rotation.

Issue #735, opened 2025-01-07
Button


halimahLDN:
- [ ] Screen reader dictates text in button twice (only when the button contains nested elements - not sure if it's desirable to change this)
- [x] Icon buttons should have a title, aria-label or similar. This is not the case in the following stories:
   - Form > FormBackButton > Customise Button
   - Form > FormSubmitButton > Customise Button
   - Solution in #790

- [x] Insufficient colour contrast. Expected 4.5:1
<img width="475" alt="image" src="https://github.com/user-attachments/assets/8eb963ba-9d4c-4f46-a14f-817caa1abb16" />
   
   - Default Primary
   - Solid Primary
   - Brand (except Default condition)
   - Text (except Secondary emphasis)

With Link, opening in new tab:
- [x] Links that open in a new tab or window [should be identified](https://www.w3.org/WAI/WCAG22/Techniques/general/G201). #782 

- [x] It's not especially clear when a Primary Outline button is in focus using keyboard tabbing.
https://github.com/user-attachments/assets/ab84a61a-60cb-4a87-9218-58433f9fe815
- Not in focus: 
![image](https://github.com/user-attachments/assets/856eea63-d917-4380-9a9d-734d6c906dde)
- In focus:
![image](https://github.com/user-attachments/assets/27bd172e-f904-4d84-9a6c-a800c4f681fb)

- [x] Text buttons should be visually distinct from text to indicate it's possible to interact with them
  - e.g. underlined
  - For example, text buttons with secondary emphasis are virtually indistinguishable from normal text:
    ![image](https://github.com/user-attachments/assets/5481a3be-df85-4cd0-b672-3d350abdafd3)

**Usability:**
Although `disabled` is sufficient to meet WCAG, [this article](https://css-tricks.com/making-disabled-buttons-more-inclusive/) lists limitations of this. Based on this, for improved usability, I suggest:
- [x] Disabled button is declared `dimmed` by screen reader but should still be locatable on the page by tabbing
   - [x] Use error messages to flag that it is disabled rather than the disabled attribute
   - [x] Replace `disabled` with `aria-disabled="true"`

ChrisKnightLDN:
related: #451

PR #782, opened 2025-01-14
Identify button links that open in a new tab

halimahLDN:
Added two options for indicating that a link opens in a new tab:

- Icon with title (so screen readers can notify the user of the functionality)
- Button with plain text so all readers know what the expected functionality of the button is

**Related issues**: #735

**Is it complete?**


halimahLDN:
Based on feedback from the team at stand-up, I have updated this to use a story to demonstrate using an icon and removed the hard-coded `#if` block adding an icon, to allow customisability.

I have also added a plain-text variation of the button with indication that the link opens in a new tab.

image

Issue #737, opened 2025-01-08
Callout


halimahLDN:
Callout - Sizes - Notice:
- [x] Landmarks should have a unique role or role/label/title (i.e. accessible name) combination. [Link](https://dequeuniversity.com/rules/axe/4.10/landmark-unique?application=axeAPI)
   - Solution proposed in #783

PR #783, opened 2025-01-14
Adding aria label to Callout

halimahLDN:
**What does this change?**
Heading now has an aria label, so screen readers can tell multiple Callout components apart in the case there's more than one on the page.

**Related issues**: #737

**Is it complete?**


.changeset/cyan-monkeys-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+ADDED: ARIA label for accessibility
```
jamesscottbrown, line 5:
```suggestion
ADDED: ARIA label to `Callout` component for accessibility
```

Issue #762, opened 2025-01-13
ObservablePlot


halimahLDN:
- [x] Elements must only use permitted ARIA attributes:
![image](https://github.com/user-attachments/assets/60a5b7b6-3645-42b6-8976-5bf94215a13b)
![image](https://github.com/user-attachments/assets/1b37d11b-a62b-41ab-a375-6e13fc34f9a3)
![image](https://github.com/user-attachments/assets/285f78d5-9e05-48cb-8020-ece6147d8957)

- [x] Elements must meet minimum colour contrast ratio thresholds
![image](https://github.com/user-attachments/assets/ff7bf2f0-6898-4e84-90b0-2da3763c24ea)
![image](https://github.com/user-attachments/assets/2811b6ad-7650-4645-8d58-c00abc6e2243)

- [x] Screen reader reads chart labels in an odd order which makes it difficult to interpret (it doesn't describe the placement of marks and reads tick labels before axis labels)
- [x] Don't use underscores or other unnecessary symbols in axis labels
![image](https://github.com/user-attachments/assets/2083d0b0-3725-49e3-9256-56cd711d6a1f)
- [x] Axis rules are read as `rule, empty group`
- [x] Invisible tick labels are still read by the screen reader
- [x] Bar and point marks are read as `graphic symbol` by screen reader 
- [x] Custom click interaction doesn't allow keyboard input
- [x] Tooltips don't appear on tabbing
- [x] Screen reader reads years and percentages in separate groups which makes it impossible to understand the data (i.e. reads '2014, 2019, 2020, 2022' and then '44%, 42%, 35%, 38%')
![image](https://github.com/user-attachments/assets/0cedd936-0a1f-4295-a82a-8b7d9bd71387)
- [x] Legend colour swatch is described as 'image' by screen reader (but Legend components don't have this issue)

halimahLDN:
Resolved by #789 which hides the SVG element and provides a separate description

PR #789, opened 2025-01-16
Added ariaHidden to exampleLineCharts

halimahLDN:
**What does this change?**
**Originally**
I've added `ariaLabel` and `ariaHidden` attributes to the example line chart, to demonstrate how we can incorporate accessibility best practice.

I don't think this is quite perfect due to the following:
- Adding ariaHidden: 'true' to Plot marks, successfully hides the SVG elements from the screen reader. Except for the axis marks, if they have a label.
- `ariaDescription` doesn’t appear to work inside marks or the spec - the screen reader ignores it
- `ariaLabel` on the `Plot.line()` makes the line disappear
- `ariaLabel` on the spec (outside of marks) DOES work but the screen reader still reads the y axis label, as it's not hidden
- The `ariaLabel` isn't achieving much more than the subTitle is, so we could instead replace it with a slightly more descriptive subTitle and hide the chart (though how to hide the chart remains to be seen, given the above re axis labels)

**Now**
I've instead added an optional prop `ariaHidden` to `ObservablePlotInner` which defaults to true, hiding the chart contents. I've removed the individual ARIA attributes in the exampleLineChart marks and updated the title and subtitle to be more descriptive.

I've since also added `ariaDetails` and `ariaDescribedBy` props which default to empty strings. So more detailed descriptions of the chart contents can be shared.

**Related issues**: #762

**Is it complete?**

### DIFF
--- a/.changeset/chilled-rabbits-shake.md
+++ b/.changeset/chilled-rabbits-shake.md
@@ -1,0 +1,6 @@
+---
+'@ldn-viz/charts': minor
+'@ldn-viz/ui': minor
+---
+
+ADDED: chartDescription modal to Footer for use in chart and chartDescription for accessibility to ChartContainer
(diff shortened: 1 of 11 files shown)


halimahLDN:
I've now implemented `chartDescription` in the example line chart, both as a modal for sighted users and an `sr-only` description in the ChartContainer.

Issue #771, opened 2025-01-13
Checkbox


halimahLDN:
- [x] Checkbox label is declared twice by screen reader
   - Once when checkbox is highlighted and once when label is highlighted
   - This doesn't happen in [this example](https://a11y-101.com/development/checkboxes)
  Solution proposed in #792

~- [ ] Disabled checkbox isn't declared by screen reader correctly (described as `dimmed unticked`)~

PR #792, opened 2025-01-16
Used semantic html to stop screen reader dictating labels twice

halimahLDN:
**What does this change?**
The checkbox was previously wrapped by a `label` element but the `span` element nested inside the `label` was confusing the screen reader, so it read the label twice.

Instead, I've wrapped everything in a `div`, replaced the `span` with a `label` element and added a `for={id}` attribute to link it to the input.

**Related issues**: #771

**Is it complete?**


jamesscottbrown:
This should have a changeset file.

PR #808, opened 2025-01-24
add colorToRGBArray function

jamesscottbrown:
This adds the `colorToRGBArray` function. This is needed in most projects that use Deck.gl, since Deck.gl requires colors to specified as an array of R/G/B values, and this is not the format in which our color tokens are defined (or which is returend by our color scales).


jamesscottbrown:
> We could update our build pipeline to supply tokens defined as rgb values (or oklch or ...) maybe this function is a practical solution for now though?

That would help when assigning something to be, for example, our primary categorical color. But it wouldn't help for when we're applying a continuous color scale, as things like the D3 color scales don't directly give colors in this format.

PR #823, opened 2025-01-29
update colors in ColorPicker when theme is switched

jamesscottbrown:


jamesscottbrown:
I'm also not sure of the best order in which to present the list of colors.

PR #829, opened 2025-01-31
align the icons in LayerControls when different sets of controls enabled

jamesscottbrown:
This fixes the alignment of controls so that all controls of the same type (e.g., opacity) are in the same column. I think it is a definite improvement, but necessarily the best final design.

image

However, it potentially looks quite gappy. Adding control buttons that look disabled might cause confusion about why they are disabled, and whether the user can do anything to re-enable them. A disabled color control that is greyed out might be confused with a color control that is set to grey (and something more explicit like a color control that is crossed-out might be confused for a control that is set to "no fill color").

@halimahLDN  also thought that a large gap caused by a series of missing controls between the checkbox and the labels might make it harder to read which checkbox corresponds to which label. @mikeldn suggested zebra-striping could help. A hover state might also help, though would be less effective on mobile. Re-ordering the checkboxes so that the checkbox was beside the label would help, but is not ideal as clicking on the checkbox to toggle whether a layer is enabled is the primary action.

@ChrisKnightLDN, this probably needs your input.

### DIFF
--- a/.changeset/align-layer-controls.md
+++ b/.changeset/align-layer-controls.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
+FIXED: align the icons in `LayerControl` instances within a `LayerControlGroup` with different controls enabled
--- a/packages/ui/src/lib/layerControl/ColorPicker.svelte
+++ b/packages/ui/src/lib/layerControl/ColorPicker.svelte
@@ -4,6 +4,8 @@
 
 	export let color = '#EE266D';
 
+	export let hideControl = false;
+
 	const colors = [
 		'#EE266D',
 		'#00AEEF',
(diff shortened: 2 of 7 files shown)


ChrisKnightLDN:
I don't think this is an improvement, and the current flow is as it should be. Unless I am missing the motivation behind this
Screenshot 2025-02-03 at 08 55 10

mikeldn:
@ChrisKnightLDN I. think the real issue that this solves is when you 'don't' have one or more of the first controls (like color) for a particular layer... then everything would align left, taking the opacity control (for example) out of alignment with the opacity controls above... 

Maybe it's OK enough "as is",  but worth discussion I think.

image

ChrisKnightLDN:
My initial thought is that changing the ux paradigm on a control by control basis is bad. if we allow one layer to have control for color, opacity and size I'd think very hard about the use case where not allowing the same for all layers (especially within a group) is correct. Lets discuss the use case that was the trigger for this change.

jamesscottbrown:
> I'd think very hard about the use case where not allowing the same for all layers (especially within a group) is correct.

It is definitely the case that we want to allow the size or color control for some layers but not others. For example, size controls can be controlled to layers that symbolize features by points, but not to layers that represent things as polygons correpsonding to their actual spatial extent. Color controls can be applied to things like points or polygons that we are drwing, but not to a satelltie imagery layer.

jamesscottbrown:
> I don't think this is an improvement, and the current flow is as it should be. Unless I am missing the motivation behind this

This is a better illustration of the problem:
image

ChrisKnightLDN:
Ok - I think then that the least evil way to achieve this is to keep the controls in place, but to visually distinguish them as disabled. If the 'greyed-out' approach used for buttons etc is not distinguishable enough we'll come up with a more explicit solution. I'll take a look, while still advising to minimise these occurrences as much as possible

ChrisKnightLDN:
closing in favor of #847

PR #833, opened 2025-02-03
Broken barchart

jamesscottbrown:
This PR should not be merged - it just adds an example of a bar chart that is broken by our current default styling.

https://dev.ldn-gis.co.uk/storybook-broken-barchart/?path=/story/charts-observableplot--bar-chart


jamesscottbrown:
This can be fixed by explicitly adding a `Plot.axisY()` mark (which then gets our default `...theme.defaultYAxis` styling applied). If the mark is not explicitly added, the mark is added by observable plot, using its (rather than our) styling.

PR #903, opened 2025-03-06
don't set color scales to ordinal type as default

jamesscottbrown:
Observale Plot should determine the scale type automatically. Setting it to `ordinal` by default will mean that it will need to be explicitly set if a different scale type is required (otherwise, instead of being rendered as a nice interpretabl color-ramp, teh legend for a continuous color scale will be drawn as a large number of color chips, one per unique value in the data).


mikeldn:
This makes sense, but having set this as the default, we'd need to review the impact on existing charts/docs.

PR #913, opened 2025-03-19
ListMenu Component

### LINKED ISSUE #911: ListMenu Component
# User Stories

**As a developer, I want to build a list component that…**

- [x]  Is one level
- [x]  Or has expandable sections with headings that…
    - [x]  Are links
    - [x]  Aren’t links
- [x] Can be set to always expanded
   Image
- [x]  Can be oriented vertically (default)
- [x]  Can be oriented horizontally

**As a mouse user, I want to…**

- [x]  Select a page using mouse
- [x]  Identify sections easily
- [x]  Expand/collapse sections without navigating to another page
- [x]  See which page is currently selected/active

**As a keyboard user, I want to…**

- [x]  Navigate the list using ~arrow keys and~ `Tab` / `Shift+Tab`
- [x]  Select a page using `Spacebar` or `Enter`
- [ ]  ~Collapse sections using `Esc`~

**As a screen reader user, I want to…**

- [x]  Navigate the list using arrow keys
- [x]  Know whether a section is expanded or collapsed
- [x]  Know which page is selected/active
- [x]  Know how many links/list items are in a section

# Acceptance Criteria

- [x]  Customisable single or multi-level hierarchy list
- [x]  List is flexible enough to be oriented vertically or horizontally
- [x]  Expandable sections are accessible with keyboard/mouse and expanding/collapsing doesn’t change page context
- [x]  Focus state is visible for keyboard users
- [x]  Active state is visible
- [x]  States are accessible to screen reader users

halimahLDN:
**What does this change?**
Implemented new `ListMenu` and `ListMenuItem` components, which is a multi-level list that is primarily intended to be used in sidebars but is flexible enough to go anywhere required.

**Why?**
We need this for our content site and the Hub and potentially other apps, so it was important to make a generic and accessible component for these and future use cases.

The content site was using Melt but didn't have expected functionality as trying to collapse lists changed context.

**Related issues**: Closes #911 

No

**How is it tested?**
Storybook

**How is it documented?**
Storybook

N/A

**Is it complete?**
No, it needs styling, which @ChrisKnightLDN will address.


packages/ui/src/lib/listMenu/ListMenuItem.svelte
```diff
@@ -0,0 +1,187 @@
+<script context="module" lang="ts">
+	export interface ListMenuItem {
+		title: string;
+		id: string;
+		url: string;
+		children?: ListMenuItem[];
+		isExpanded?: boolean;
+	}
```
jamesscottbrown, line 8: It's not ideal that the name of the `ListMenuItem` component is the same as the type of the elements in one of its props. This requires rename the type when importing into `ListMenu`, which then results in seeing `export let items: Item[];` in `ListMenu` but `export let children: ListMenuItem[] = [];` in the `ListMenuItem` component - it isn't imemdiately obvious that these are the same type. I'm not sure of the best name, but `ListMenuEntry` for the type would work.

packages/ui/src/lib/listMenu/ListMenu.svelte
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * Optional prop to change orientation
+	 */
+	export let orientation: 'vertical' | 'horizontal' = 'vertical';
+
+	/**
+	 * Exposes active menu item to parent container for modification.
+	 */
+	export let selectedMenuItemId = writable('');
+
```
jamesscottbrown, line 43: Passing stores as props to components can be inconvenient - is there a reason why this prop is a `Writable<string>` rather than just a string?

halimahLDN: Thanks, James, that makes sense. Chris and I wrote this together and wanted to expose a prop that could be assigned to `$selectedValue`. However, it got mixed up and we ended up modifying `$selectedMenuItemId` directly which required it to be a store. It makes sense to make `selectedMenuItemId` a string as you say and replace anything that uses `$selectedMenuItemId` with `$selectedValue`.

packages/ui/src/lib/listMenu/ListMenu.svelte
```diff
@@ -0,0 +1,104 @@
...
+						mapItems(child.children, targetId).some((c: ListMenuItem) => c.isExpanded))
+			);
+
+			return {
+				...item,
+				isExpanded: hasMatchingChild || item.id === targetId,
+				children: item.children ? mapItems(item.children, targetId) : undefined
+			};
+		});
+	};
```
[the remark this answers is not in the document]
halimahLDN: Thank you - I've made this change!

packages/ui/src/lib/listMenu/ListMenuItem.svelte
```diff
@@ -0,0 +1,187 @@
...
+		isExpanded = !isExpanded;
+	};
+
+	/**
+	 * When the parent list collapses, any children that are expanded also collapse.
+	 * This means next time the parent list is expanded, the children aren't already expanded.
+	 */
+	const toggleChildren = (children: ListMenuItem[], expanded: boolean) => {
+		return children.map((child) => {
+			if (expanded === false && child.isExpanded === true) {
```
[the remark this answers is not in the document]
halimahLDN: I looked at this but I wasn't sure how to handle passing around the `isExpanded` state, so I haven't implemented it. If you are able to implement it, please could you explain how you've done it?

PR #933, opened 2025-04-24
maps: add MapArrow component

jamesscottbrown:
**What does this change?**

This adds a `MapArrow` component, to superimpose an arrow pointing at a point on a map.

**Why?**

On the Green Infrastructure map, individual hexes are small, so highlighting their boder is not enough to enable users to quickly find them: adding an arrow helps.

No.

**Is it complete?**

### DIFF
--- a/.changeset/map-arrow.md
+++ b/.changeset/map-arrow.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/maps': minor
+---
+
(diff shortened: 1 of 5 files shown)


jamesscottbrown:
Preview: https://dev.ldn-gis.co.uk/storybook-map-arrow/?path=/docs/maps-components-maparrow-maparrow--documentation

PR #937, opened 2025-04-28
tables: improve ARIA

jamesscottbrown:
This improves the ARIA for the table component.

For the [basic example table](https://dev.ldn-gis.co.uk/storybook-table-aria/?path=/story/tables-components-table--default), Vocie Assitant allows navigation with Control-Option-[up/down/left/right] and correctly describes cell contents, e.g.:

* "First Name Margharet. Column 1 of 3" - when shifting between columns
* "Row 14 of 21: Margharet" - when shifting between rows

### DIFF
--- a/packages/tables/src/lib/core/renderers/Header.svelte
+++ b/packages/tables/src/lib/core/renderers/Header.svelte
@@ -46,13 +46,7 @@
 	$$restProps;
 </script>
 
-<div
-	on:click={toggle}
-	on:keypress={toggle}
-	role="cell"
-	tabindex={0}
-	class="font-bold py-0.5 w-full h-full"
->
+<div on:click={toggle} on:keypress={toggle} tabindex={0} class="font-bold py-0.5 w-full h-full">
 	<div class={classNames('flex items-center select-none', alignmentClass)}>
 		{#if superscriptText}
 			<div class="text-left">
--- a/packages/tables/src/lib/core/renderers/Header.svelte
+++ b/packages/tables/src/lib/core/renderers/Header.svelte
@@ -46,10 +46,11 @@
 	$$restProps;
 </script>
 
-<div on:click={toggle} on:keypress={toggle} tabindex={0} class="font-bold py-0.5 w-full h-full">
+<div class="font-bold py-0.5 w-full h-full">
 	<svelte:element
 		this={allowSorting ? 'button' : 'div'}
 		class={classNames('flex items-center select-none', alignmentClass)}
+		on:click={allowSorting ? toggle : undefined}
 	>
 		{#if superscriptText}
 			<div class="text-left">
--- a/packages/tables/src/lib/core/renderers/Header.svelte
+++ b/packages/tables/src/lib/core/renderers/Header.svelte
@@ -47,7 +47,10 @@
 </script>
 
 <div on:click={toggle} on:keypress={toggle} tabindex={0} class="font-bold py-0.5 w-full h-full">
-	<div class={classNames('flex items-center select-none', alignmentClass)}>
+	<svelte:element
+		this={allowSorting ? 'button' : 'div'}
+		class={classNames('flex items-center select-none', alignmentClass)}
+	>
 		{#if superscriptText}
 			<div class="text-left">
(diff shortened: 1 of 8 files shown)


halimahLDN (changes requested):
Hello, I've looked over this and picked up on a few further accessibility issues:
- there's an invisible row below the column headers, which causes the screen reader to incorrectly state there are 21 rows (but there are actually only 20 rows visible)
image

- According to the [Design Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/), the `<div role="row">` surrounding the column headers should be nested inside a `<div role="rowgroup">`. Similarly all of the rows that make up the table body should be nested inside a `<div role="rowgroup">`.
- If possible, it would be ideal to implement the [pattern for sortable tables](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) as described here. In particular:
    - When the table headers are sortable they are clickable but this isn't immediately obvious because the elements are divs, not buttons, and sr users won't know about the sorting at all. These should be buttons when `allowSorting="true"` with visual keyboard focus, hover state and cursor pointer. 
    - add a caption to the table, with an `sr-only` note that column headers with buttons are sortable. (perhaps this could be an `ariaLabel` if no visible caption?)
    - add an `aria-sort` attribute to the column header that's being sorted, which tells screen reader users what's being sorted and whether it's ascending or descending

packages/tables/src/lib/core/renderers/Header.svelte
```diff
@@ -47,7 +47,10 @@
 </script>
 
 <div on:click={toggle} on:keypress={toggle} tabindex={0} class="font-bold py-0.5 w-full h-full">
```
halimahLDN, line 49: I think this div wrapping the button is causing some odd behaviour, probably because it's got an `on:click`, `on:keypress` and `tabindex` that's interfering with the button functionality. When I'm using `Tab` with one of the elements focused, I can use spacebar or enter to toggle the sorting but if I tab again, I'm unable to use Enter and spacebar looks like it's toggling but resets back to the previous setting.

https://github.com/user-attachments/assets/021bc5af-aac7-4b2a-a29c-143368f77e62

Is it possible to remove the event listeners and tabindex from this div (or do we even need this div)?

packages/tables/src/lib/table/rows/headerRows/ColumnHeadingRow.svelte
```diff
@@ -42,12 +42,19 @@
...
 			{#if !table.visibleFields || table.visibleFields.includes(col.short_label)}
-				<div class="flex was-th" role="columnheader" style:width={col.computedWidth + 'px'}>
+				{@const colIsSortable = allowSorting && col.sortable !== false}
+				{@const order = table.rowOrderSpec.find((f) => f.field === col.short_label)?.direction}
+
+				<div
+					class="flex was-th"
+					role="columnheader"
+					style:width={col.computedWidth + 'px'}
+					aria-sort={colIsSortable ? order : ''}
```
halimahLDN, line 52: 'sort up' and 'sort down' doesn't mean much to screen reader users. Is it possible to make this more descriptive, i.e. `ascending/descending`?

Also, is it possible to add an ariaLabel or something to the table (not individual buttons) telling screen reader users that the column buttons are there for sorting the data?

halimahLDN (commented):
Just a few more things I've picked up on. Let me know if anything is unclear.

packages/tables/src/lib/table/rows/headerRows/ColumnHeadingRow.svelte
jamesscottbrown: > 'sort up' and 'sort down' doesn't mean much to screen reader users. Is it possible to make this more descriptive, i.e. ascending/descending?

I don't understand the question: `aria-sort` is being set to `ascending` or `descending`.

halimahLDN: Sorry, for clarity when the screen reader narrates the sort, it says 'sort up' and 'sort down' instead of 'sort ascending'/'sort descending'. I assumed that's because of the value of `order` but perhaps not?

PR #943, opened 2025-04-29
A11y: updated LayerControl and LayerControlGroup to apply name to checkboxes

halimahLDN:
**What does this change?**
- `LayerControlGroup` now passes `name` into `LayerControl`
- `LayerControl` now passes `name` into `Checkbox`, in addition to `Radio`
- `LayerControlGroup` and `CheckboxGroup` now have an `aria-controls` attribute on the checkbox that toggles all controls, with a value equal to the ids for all controls

**Why?**
For better accessibility

No

**How is it tested?**
Storybook

**How is it documented?**
Storybook

N/A

**Is it complete?**

### DIFF
--- a/.changeset/brown-glasses-shop.md
+++ b/.changeset/brown-glasses-shop.md
@@ -2,8 +2,8 @@
 '@ldn-viz/ui': minor
 ---
 
-CHANGED: `CheckboxGroup`, `LayerControl` and `LayerControlGroup` components for better accessibility
+CHANGED: `CheckboxGroup`, `LayerControl` and `LayerControlGroup` components
 
 - `LayerControlGroup` now passes `name` into `LayerControl`
 - `LayerControl` now passes `name` into `Checkbox`, in addition to `Radio`
-- `LayerControlGroup` and `CheckboxGroup` now have an `aria-controls` attribute on the checkbox that toggles all controls, with a value equal to the ids for all controls
+- For accessibility: `LayerControlGroup` and `CheckboxGroup` now have an `aria-controls` attribute on the checkbox that toggles all controls, with a value equal to the ids for all controls
--- a/.changeset/brown-glasses-shop.md
+++ b/.changeset/brown-glasses-shop.md
@@ -2,7 +2,8 @@
 '@ldn-viz/ui': minor
 ---
 
-CHANGED: applied `name` attribute to `Checkbox` components within `LayerControl` and `LayerControlGroup` for better accessibility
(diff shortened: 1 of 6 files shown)


packages/ui/src/lib/layerControl/LayerControl.stories.svelte
jamesscottbrown, line 1: Since the `name ` prop is optional, I would remove it from most of these stories and have a separate story that demonstrates it.

packages/ui/src/lib/layerControl/LayerContolGroup.stories.svelte
```diff
@@ -10,25 +10,28 @@
...
 			id: 'train',
+			name: 'transport',
 			label: 'Train stations',
 			hint: 'Excluding underground stations'
 		},
-		{ id: 'underground', label: 'Underground stations' },
-		{ id: 'taxi', label: 'Taxi ranks', disabled: true }
+		{ id: 'underground', name: 'transport', label: 'Underground stations' },
+		{ id: 'taxi', name: 'transport', label: 'Taxi ranks', disabled: true }
 	];
```
jamesscottbrown, line 22: If this is for mutually exclusive layers, the `name` should be set from the `name` prop of the `LayerControlGroup` rather than needing to be separately set for each option. If they're not mutually exclusive, then the `name` for the controls for different layers shouldn't be set to the same value. In either case I'd remove the `name` proprerties from here.

.changeset/brown-glasses-shop.md
```diff
@@ -0,0 +1,9 @@
+---
+'@ldn-viz/ui': minor
+---
+
+CHANGED: `CheckboxGroup`, `LayerControl` and `LayerControlGroup` components for better accessibility
+
+- `LayerControlGroup` now passes `name` into `LayerControl`
+- `LayerControl` now passes `name` into `Checkbox`, in addition to `Radio`
```
jamesscottbrown, line 8: I'm not sure these changes actually improve accessaibility.

The `name` attribute seems to be ignored by screenreaders, which instead announce the checkbox using the `label`. It's used to determine the name given to the value of an inpt control in a form submission, and to tell the broser which radio buttons belong to the same group, so has a role like more like an inernal `id` than a label that's intended to be meaningful to users.

halimahLDN: Yes, I wasn't sure about this either as the design pattern doesn't include the name attribute but it was flagged as part of the accessibility review of the Voter ID map, which is why I made this PR.

PR #960, opened 2025-05-08
ColorPicker accepts custom colours

### LINKED ISSUE #959: layerControl to allow use of custom set of color tokens
[layerControl
](https://github.com/Greater-London-Authority/ldn-viz-tools/tree/dev/packages/ui/src/lib/layerControl) currently uses the standard categorical (light & dark) colors.

In some cases it would be useful to add to / replace the colours with custom colors.

Each color added needs a light & dark equivalent, so color tokens like color-palette-green-300 would make sense.

It would be ideal to:
1. No colors array passed in? : Use default categorical (as now)
2. Colors array passed in? : Use that list only, in order defined

We should be able to use a mix of categorical and palette tokens, like:

[color-palette-green-300, color-palette-pink-800, color-data-categorical-red, etc]

halimahLDN:
**What does this change?**
`LayerControlGroup`, `LayerControl` and `ColorPicker` accept a custom list of colour tokens, called `colorNames`.

**Why?**
This allows for use in maps where specific colours are required (e.g. requiring a light blue for water).

**How?**
`colorNames` is initialised as an empty array. If an array of token strings is passed in, it will show those colours. Otherwise, if it's empty, it will default to standard categorical colours, as normal.

image

image

**Related issues**: Closes #959 

No

**How is it tested?**
Storybook

**How is it documented?**
Storybook

Yes (colorNames are token names, so change depending on the theme)

**Is it complete?**

### DIFF
--- a/packages/ui/src/lib/layerControl/ColorPicker.svelte
+++ b/packages/ui/src/lib/layerControl/ColorPicker.svelte
@@ -13,7 +13,12 @@
 
 	export let disabled = false;
 
-	const colorNames = [
+	/**
+	 * Optional custom colours to choose from. If these don't exist, default to categoricalColors.
+	 */
+	export let colorNames;
+
+	const categoricalColors = [
 		'data.categorical.blue',
 		'data.categorical.darkpink',
 		'data.categorical.pink',
@@ -26,6 +31,10 @@
 		'data.neutral.0',
 		'data.neutral.1'
 	];
+
+	if (!colorNames.length) {
+		colorNames = categoricalColors;
+	}
 
 	let isOpen = false;
 </script>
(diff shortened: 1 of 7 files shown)


packages/ui/src/lib/layerControl/ColorPicker.svelte
```diff
@@ -13,7 +13,12 @@
 
 	export let disabled = false;
 
-	const colorNames = [
+	/**
```
jamesscottbrown, line 16: Why not

```js
	export let colorNames = [
		'data.categorical.blue',
		'data.categorical.darkpink',
		'data.categorical.pink',
		'data.categorical.red',
		'data.categorical.yellow',
		'data.categorical.green',
		'data.categorical.purple',
		'data.categorical.orange',
		'data.categorical.turquoise',
		'data.neutral.0',
		'data.neutral.1'
	];
```

PR #967, opened 2025-05-15
A11y: Added ColorPicker tooltip, hover and active states

### LINKED ISSUE #961: A11y: ColorPicker
**Is your feature request related to a problem? Please describe.**
When the `ColorPicker` opens, it's not immediately obvious to all users which colour is already selected. Also, for users with colour deficiency, it would be helpful to identify the colours on hover. Moreover, for all users, it is confusing that there isn't a hover state when you move the cursor over the buttons.

**Describe the solution you'd like**

- Highlight the currently selected colour when the `ColorPicker` opens
- Add a tooltip that identifies the focused colour on hover
- Add a hover state for sighted users to identify which button is currently in focus (to match keyboard focus state)

**Describe alternatives you've considered**
The `ColorPicker` does already have ARIA labels but these are only useful for screen reader users.

**Additional context**
Keyboard focus:

Image

halimahLDN:
**What does this change?**
`ColorPicker` now has the following:
- Each colour option has a tooltip with the name of the colour, for better accessibility
- When you hover over a colour, a ring appears
- Active colour (currently selected) has a ring around it 

**Why?**
Improve accessibility

**Related issues**: Resolves #961 

No

**How is it tested?**
Storybook

**How is it documented?**
Storybook

Yes

**Is it complete?**
Requires design review of focus state as this is currently not visible.

### DIFF
--- a/packages/ui/src/lib/layerControl/ColorPicker.svelte
+++ b/packages/ui/src/lib/layerControl/ColorPicker.svelte
@@ -10,7 +10,7 @@
 
 	export let label;
 
-	export let colorName = 'data.categorical.blue';
+	export let activeColorName = 'data.categorical.blue';
 
 	export let disabled = false;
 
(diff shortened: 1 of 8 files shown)


halimahLDN:
Last two changes were to add the `ring-offset` colour, so the active colour has a visible ring around it and to fix the default `hover:bg` from `Button.svelte` coming through. See difference below:

Before changes (the darkPink button has a hover bg different to the container colour which makes the ring look a bit square):
image

After changes:
image

packages/ui/src/lib/layerControl/ColorPicker.svelte
```diff
@@ -37,6 +38,18 @@
...
+		const array = token.split('.');
+		array.shift();
+		return array.join(' ');
+	};
+
+	const activeOptionClasses =
+		'ring-inset ring-2 ring-offset-2 ring-color-ui-background-secondary ring-offset-color-ui-border-primary hover:ring-offset-color-ui-border-primary focus-visible:ring-offset-color-ui-border-primary';
+
+	const optionClasses =
+		'rounded-full bg-color-container-level-0 hover:bg-color-container-level-0 hover:ring-inset hover:ring-offset-2 hover:ring-offset-color-ui-border-secondary hover:ring-2 hover:ring-color-ui-background-secondary focus-visible:ring-offset-color-ui-border-secondary';
```
halimahLDN, line 52: @ChrisKnightLDN  is `ring-color-ui-background-secondary` (in activeOptionClasses and with `hover` modifier in optionClasses) intentional to differentiate between focus and hover states? The focus has a `ring-color-ui-background-primary`, which I think looks better

PR #988, opened 2025-06-12
Sv5 overlay components

ChrisKnightLDN:
**What does this change?**
Adds overlay componets refactored to svelte 5 using bits ui

### DIFF
--- a/packages/ui/src/lib/overlay/Overlay.svelte
+++ b/packages/ui/src/lib/overlay/Overlay.svelte
@@ -1,0 +1,108 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	/**
+	 * The `<Overlay>` component provides additional explanatory or help text when a user interacts with a trigger.
+	 * You can choose whether this is a modal, popover or tooltip, depending on your needs.
+	 *
+	 * The trigger can either be the built in `<Trigger>` component or a custom trigger passed into the relevant slot.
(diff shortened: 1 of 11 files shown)


packages/ui/src/lib/overlay/Overlay.svelte
```diff
@@ -0,0 +1,91 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	/**
+	 * The `<Overlay>` component provides additional explanatory or help text when a user interacts with a trigger.
+	 * You can choose whether this is a modal, popover or tooltip, depending on your needs.
+	 *
+	 * The trigger can either be the built in `<Trigger>` component or a custom trigger passed into the relevant slot.
```
jamesscottbrown, line 7:
```suggestion
	 * The trigger can either be the built in `<Trigger>` component or a custom trigger provided as a snippet.
```

ChrisKnightLDN:
> The default tooltip [delayDuration](https://www.bits-ui.com/docs/components/tooltip#delay-duration) feels far too long.

agree, we had ```openDelay: 0,
		closeDelay: 0,
		closeOnPointerDown: false,
		disableHoverableContent: true``` in the old setup with a ```transition:fade={{ duration: 100 }}``` on the content element

jamesscottbrown:
This seems fine now.

PR #1009, opened 2025-07-10
Sv5 checkbox group

ChrisKnightLDN:
**What does this change?**
adds checkbox group


packages/ui/src/lib/checkBox/CheckBoxGroupDemo.svelte
```diff
@@ -0,0 +1,43 @@
+<script lang="ts">
```
jamesscottbrown, line 1:
```suggestion
// N.N. This story has been extracted into a separate component because each Story is wrapped by a snippet; this would capture the `customOverlay()` snippet, preventing it from being accessed in the `<script>` section and added to the  objects in the `optionsWithCustomOverlay` array.

<script lang="ts">
```

PR #1011, opened 2025-07-15
add async button + loading indicator and spinner

ChrisKnightLDN:
**What does this change?**
adds async button, and required loading indicator and spinner


jamesscottbrown:
The autodocs properties table is missing the descriptions of several props (e.g. `variant`) - this is because the docstrings for these need to be ont he *type* of the prop, rather than being listed when `$props` are being destructured.

Rather than 

```
	let {
		/**
		 * Selects which family of styles should be applied to the button.
		 */
		variant = 'solid',
...
		...restProps
	}: ButtonProps = $props();
```

in `Buttons.svelte`, there should be

```js
type LdnButtonProps = {
	/**
        * Determines the visual emphasis is placed on the button.
       */
	variant?: 'brand' | 'square' | 'solid' | 'outline' | 'text';
...
};
```

in `button/types.ts`.

Making this change will add the description to the table for both the `Button` and `AsyncButton` components.

jamesscottbrown:
Also, the properties table lists both `on:submit` and `onsubmit` as separate props (and similarly for other events).

packages/ui/src/lib/button/AsyncButton.svelte
```diff
@@ -0,0 +1,133 @@
...
+	// /**
+	//  * Sets the size of the button.
+	//  */
+	// export let size: ButtonProps['size'] = 'md';
+
+	// /**
+	//  * If `true`, then the button cannot be interacted with (either by clicking,
+	//  * or by using the keyboard).
+	//  */
+	// export let disabled: ButtonProps['disabled'] = false;
```
jamesscottbrown, line 77:
```suggestion
```

PR #1012, opened 2025-07-15
Convert maps to Svelte 5

jamesscottbrown:


jamesscottbrown:
This is mostly done, but the geocoder will giver errors until `Geocoder, GeocoderSuggestionList` in `@ldn-viz/ui` are migrated.

PR #1055, opened 2025-07-25
Sv5 fix radiogroup label

halimahLDN:
**What does this change?**
Fixes regression where label isn't rendered if icons don't exist.

image


ChrisKnightLDN:
lgtm

PR #1070, opened 2025-08-01
Sv5 - Added default plot fragments

### LINKED ISSUE #882: ObservablePlot - Propose adding default stroke and fill to rect, barX & barY
**Is your feature request related to a problem? Please describe.**
House style is that coloured bars (usually stacked bars) should have a keyline seperation between. The keyline should be the current them BG color. It's annoying to have to explicitly set it.

**Describe the solution you'd like**
Add a new defaultRect and defaultBar styles.
fill: $currentTheme.color.data.primary
stroke: $currentTheme.color.chart.background

halimahLDN:
**What does this change?**
Added:
- defaultRect and defaultBar plot fragments (as per Mike's request in #882)

Changed:
- plot function now checks for `options.facet` as well as `options.fy/fx` to apply default facet styling

**Related issues**: Resolves #882 

**Is it complete?**

### DIFF
--- a/.changeset/angry-pens-sort.md
+++ b/.changeset/angry-pens-sort.md
@@ -1,0 +1,7 @@
+---
+'@ldn-viz/charts': minor
+---
+
+ADDED: defaultRect and defaultBar plot fragments
+
+CHANGED: plot function now checks for options.facet to apply defaultSizeFacet fragment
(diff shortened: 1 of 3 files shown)


mikeldn:
Looks good! Defaults applying sensible line and fill colours.
It's logical we're using color.chart.background as outline colour (as they should be in step), but you could argue (like with maps) that we need a token for color.element.outline (or something). But we don't, so this is good for now.

PR #1106, opened 2025-09-18
create sv5 MapLegend

halimahLDN:
Migrates MapLegend component to SV5


ChrisKnightLDN:
@halimahLDN can we keep the convention of default stories being "Default"? Here it is "Legend on map"

Issue #1129, opened 2025-10-01
SV5 Map flickers on window resize


halimahLDN:
**Describe the bug**
- When resizing the window, the map flickers

**To Reproduce**

1. Go to [SV5 Connectivity Map preview](https://dev.ldn-gis.co.uk/connectivity-sv5-refactor-map/) or [Storybook](https://dev.ldn-gis.co.uk/storybook-svelte5/?path=/story/maps-components-map--responsive-to-theme&globals=theme:light)
2. Resize window
3. See error

**Expected behavior**
- The map should not flicker/change when window resizes

**Screenshots**

https://github.com/user-attachments/assets/b843ee23-3ed4-423d-914e-941c67726826

**Desktop (please complete the following information):**

- OS: macOS 15.6.1
- Browser Chrome and Firefox
- Version 140.0.7339.213 and 143.0.1

PR #1136, opened 2025-10-01
SV5: MapLibre - fix map flickering on resize

halimahLDN:
**What does this change?**
Replaces side effect with onresize event listener, to resize MapLibre instance without flickering.

**Why?**
Fix flickering

**How?**

**Related issues**: #1129 

No

**How is it tested?**
[Storybook](https://dev.ldn-gis.co.uk/storybook-sv5-stopMapFlicker/?path=/story/maps-components-mapcontextlayers-boroughscontextlayer--default)
compared to [svelte5 branch](https://dev.ldn-gis.co.uk/storybook-svelte5/?path=/story/maps-components-mapcontextlayers-boroughscontextlayer--default)

**How is it documented?**
[Storybook](https://dev.ldn-gis.co.uk/storybook-sv5-stopMapFlicker/?path=/story/maps-components-mapcontextlayers-boroughscontextlayer--default)

N/A

**Is it complete?**

### DIFF
--- a/.changeset/icy-lies-relax.md
+++ b/.changeset/icy-lies-relax.md
@@ -2,4 +2,4 @@
 '@ldn-viz/maps': minor
 ---
 
-FIXED: replace side effect with onresize event listener to stop map flickering on window resize
+FIXED: update MapLibre.svelte, replacing side effect with onresize event listener to stop map flickering on window resize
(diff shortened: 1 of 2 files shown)


jamesscottbrown:
This resizes when the window resizes, rather than when the div containing the map resizes.

In most cases this is ok, but we might have layouts where a sidebar is beside the map (ratehr than suprimposed on it), such that opening or closing the sidebar needs to trigger a map resize. IIRC, at one point there was a `sidebarPush` option on the AppLayout to enable these layouts.

PR #1167, opened 2025-11-10
allow MapDraw component to load GeoJSON files with excessive coordinate precision 

jamesscottbrown:


jamesscottbrown:
This change has been made to the maps package in a different commit.

PR #1290, opened 2026-02-18
D3 donut

jamesscottbrown:


.changeset/every-lies-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ldn-viz/charts': minor
+---
+
+Addition of d3 donut chart component to vix tools
```
jamesscottbrown:
```suggestion
ADDED: `Donut` commponent
```

jamesscottbrown (commented):
Sometimes there isn't space for the table beside the donut, so it would be useful to have a boolean prop (e.e, `hideTable`) that allows the user to remove the `DonutLegend` component, and instead render `ColorLegendOrdinalChips` underneath the chart title.

The component currently assumes the data is an array whose elements have `label` and `value` fields. It would be good for these field names to be accepts as props (`labelField` and `countField`, which could have default values of `label` and `value`).

Eslint alos gives warnings about `#each` blocks not having keys.

PR #1351, opened 2026-07-09
replace some old color tokens with their replacement values

jamesscottbrown:
This needs review by @ChrisKnightLDN.

### DIFF
--- a/.changeset/replace-old-tokens.md
+++ b/.changeset/replace-old-tokens.md
@@ -1,0 +1,5 @@
+---
+'@ldn-viz/ui': minor
+---
+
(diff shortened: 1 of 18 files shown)


ChrisKnightLDN (changes requested):
I see only one instance where the incorrect token appears to have been applied. In breadcrumbs: the change from <span class="text-color-ui-primary"> to <span class="text-color-text"> will de-emphasise the text. The correct mapping would be to: <span class="text-color-surface-accent"> (or you could make a case for text-color-interactive-on - but they map to the same colour so it's a semantic coin toss )

